PR #248, opened 2014-03-12
Allow recursive_cmds override via REBAR_NON_RECURSIVE_CMDS

tuncer:
For the vast majority of projects it makes no sense to apply
'rebar compile' recursively, except when explicitly running
'rebar prepare-deps' or 'rebar refresh-deps'. So, to keep deps
and sub projects untouched, we need a way to disable the default
recursive behavior of 'rebar compile'.

The environment variable REBAR_NON_RECURSIVE_CMDS can be used
to pass in a comma separated list of commands to remove from the
built-in list of recursive commands. Commands ending with *-deps
are ignored.


tuncer:
Closing due to missing conclusion.

PR #273, opened 2014-05-06
Use target_dir as source of new version in generate-appups

cybergrind:
Not sure what exact reason of setting NewVerPath as rel + `first rel name from reltool.config` so trying to make this command more correct.

### DIFF
--- a/src/rebar_appups.erl
+++ b/src/rebar_appups.erl
@@ -54,7 +54,7 @@
 
     %% Get the new and old release name and versions
     {Name, _Ver} = rebar_rel_utils:get_reltool_release_info(ReltoolConfig),
-    NewVerPath = filename:join([TargetParentDir, Name]),
+    NewVerPath = rebar_rel_utils:get_target_dir(Config, ReltoolConfig),
     {NewName, NewVer} = rebar_rel_utils:get_rel_release_info(Name, NewVerPath),
     {OldName, OldVer} = rebar_rel_utils:get_rel_release_info(Name, OldVerPath),
 


lrascao:
My reltool.config only has one rel entry in sys, so it doesn't really affect my use case. The proposed fix works fine.
From what i could understand from the code (rebar_appups.erl) as it stands on master right now:
  - through rebar_rel_utils:get_target_parent_dir, target_dir is read from reltool.config, and obtains parent dir (eg. _rel/release_1 returns _rel)
  - rebar_rel_utils:get_reltool_release_info expects a single {rel, ...} tuple, it extracts the release name from it and appends it to the target parent dir

@cybergrind probably caught this because he's using multiple rel entries in the sys tuple, looking at the doc at http://www.erlang.org/doc/man/reltool.html i see this is allowed. It appears that the root of the issue is actually the fact that rebar is not allowing more than one release, the issue now is if this was an intentional design decision.
@cybergrind, do you have an actual use case that we could look at with a reltool.config and multiple rel entries?

cybergrind:
Sure

```erlang
{sys, [
       {lib_dirs, ["../deps"]},
       {erts, [{mod_cond, derived}, {app_file, strip}]},
       {app_file, strip},
       {rel, "igr", "0.9",
        [
         kernel,
         stdlib,
         sasl,
         lager,
         folsom,
         folsom_unix,
         igr_worker,
         http_api
        ]},
       {rel, "start_clean", "",
        [
         kernel,
         stdlib
        ]},
       {boot_rel, "igr"},
       {profile, embedded},
       {incl_cond, derived},
       {excl_archive_filters, [".*"]}, %% Do not archive built libs                                                                                                           
       {excl_sys_filters, ["^bin/.*", "^erts.*/bin/(dialyzer|typer)",
                           "^erts.*/(doc|info|include|lib|man|src)"]},
       {excl_app_filters, ["\.gitignore"]},
       {app, igr_worker, [{mod_cond, app}, {incl_cond, include},
                          {lib_dir, "../apps/igr_worker"}]},
       {app, http_api, [{mod_cond, app}, {incl_cond, include},
                          {lib_dir, "../apps/http_api"}]}
      ]}.

{target_dir, "igr"}.
```

My main concern that if we have `target_dir` for setup our output directory - we have to use it for all operations that have no additional settings without implicit cases (like generate-appups).

lrascao:
So i looked at the original commit and really can't understand the reason for the parent target dir + first rel name from reltool.config approach for appups, @cybergrind's change looks sane. Would be nice to get more input from people who actually use appups but appears there aren't much around.

tuncer:
ping?

lrascao:
i don't really have a lot to add other that what i've already said, so :+1: by me

PR #277, opened 2014-05-09
Run update-deps hooks again

Vagabond:
Fix regression caused by making update-deps actually work.

### DIFF
--- a/src/rebar_core.erl
+++ b/src/rebar_core.erl
@@ -28,6 +28,8 @@
 
 -export([process_commands/2, help/2]).
 
+-export([execute_pre/5, execute_post/5]).
+
 -include("rebar.hrl").
 
 %% ===================================================================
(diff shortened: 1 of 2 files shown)


tuncer:
So, what do we do about this? As I said earlier, given a clear NOTE/TODO comment, it's okay to merge, assuming this fixes the regression.

ferd:
I frankly have no idea. I don't understand this code base, don't understand the implications, never know what may break or not. I'm fine with merging it and reversing if it breaks more stuff, though.

tuncer:
So, how about merging this to master after the next release, and see if anything breaks?

ferd:
Sure, that sounds like a plan. The current release is still blocked on getting the proper protobuffs configs fixed, though.

tuncer:
I think it's safe to say this can be closed (together with #561).

PR #293, opened 2014-06-10
Check C source dependencies in needs_compile

liskin:
This makes rebar recompile/relink port drivers properly when only C headers change.

I tested this on Linux with gcc, I'm not sure whether this is portable.


tonyrog:
Wonderful! 
Mega thanks for this fix :-)
This is good progress!

PR #304, opened 2014-06-20
Fix eunit_surefire crash with rebar eunit tests=...

liskin:
This fixes the following crash:

> =ERROR REPORT==== 20-Jun-2014::16:35:38 ===
> Error in process <0.40.0> with exit value:
{function_clause,[{eunit_surefire,escape_suitename,[undefined,[]],[{file,"eunit_surefire.erl"},{line,416}]},{eunit_surefire,write_report,2,[{file,"eunit_surefire.erl"},{line,258}]},{lists,foreach,2,[{file,"lists.erl"},{line,1336}]},{...

### DIFF
--- a/src/rebar_eunit.erl
+++ b/src/rebar_eunit.erl
@@ -458,7 +458,7 @@ make_test_primitives(RawTests) ->
                             %% Generator
                             MakePrimitive(generator, M, F2)
                     end,
-                [NewFunction|Acc]
+                [eunit_module_suite(M, NewFunction)|Acc]
         end,
     lists:foldl(F, [], RawTests).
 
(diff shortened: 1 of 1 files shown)


tuncer:
Any chance of getting this finished and into the next release (to happen really soon)?

liskin:
Not much, I'm sorry. I didn't write any Erlang for almost a year now and therefore I am not a user of rebar any more. I wish I could force myself to do this for the sake of closure, but I can't promise much. :-(

lrascao:
@liskin i've cherry-picked you commit and added a test case, do you mind looking over #591 and see if the test sufficiently covers the bug you found?

liskin:
@lrascao If that test fails without my patch and succeeds with it, then I think it's okay. I can't remember the exact details of this issue, but I don't think there was more to it than a simple tests=something. :-)

lrascao:
Closing in favor of #591 as it contains the fix plus a simple test

PR #315, opened 2014-07-06
Easy name change

fgallaire:
Help to easily make projects based on rebar (like rebar + lot of specific templates).
Just change the .app file and the macro APP with the name of the new application and bootstrap will generate your project.


lrascao:
Does this issue still apply or can we close it? I guess now with rebar3's plugin architecture this doesn't make sense anymore

ferd:
Yeah I'm not sure it does either. Will close for now.

PR #317, opened 2014-07-09
Protobuffs compiler take more opts other than just erl_opts

UglyTroLL:
Basically, this issue was brought up by https://github.com/rebar/rebar/pull/150 but it failed to get merged back. I did some improvements on that pull request and Im trying to bringing it up again since I believe it is pretty important for some ppl like me.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -126,3 +126,4 @@ Yuki Ito
 alisdair sullivan
 Alexander Verbitsky
 Andras Horvath
+Zhibo Wei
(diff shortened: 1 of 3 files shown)


tuncer:
ping?

lrascao:
i too i'm not familiar with basho's protobuffs, in general i agree with tomas's input

UglyTroLL:
Actually, I tried to merge the current master with my fix, and I found that this feature has already been supported with https://github.com/rebar/rebar/commit/bf2b3e246821e5697b6154ce0d2254600c91b34c

UglyTroLL:
And the current solution is better than mine here, I believe this PR can be closed if no one has further concerns?

PR #335, opened 2014-08-07
Add flag to ignore pre- and post-hooks

derek121:
Adds the ability to specify on the command line that pre-hooks
(--ignore-pre-hooks), post-hooks (--ignore-post-hooks), or both
(--ignore-hooks) should be ignored.

This can be useful when a hook is expensive to execute, and not needed for every compile or clean. For example, having relx as a post-compile hook, to create a release, which isn't needed when doing compiles during development.

Otherwise, one's rebar.config could be manually modified to remove the hooks until they're needed again, or a copy of the rebar.config could be maintained, without the hook, but those both require extra work.

Separate makefile targets can include or not include an --ignore-* flag. For example "make dev" can invoke "rebar --ignore-post-hooks compile", while "make release" can invoke "rebar compile" (to allow the normal hook execution).

inttest/thooks_exclude has been added to test this (based on the existing inttest/thooks)

No additional Dialyzer warnings are generated from these changes ("make check" on the current master results in about 15 warnings, however).

### DIFF
--- a/inttest/thooks_exclude/fish.erl
+++ b/inttest/thooks_exclude/fish.erl
@@ -1,0 +1,5 @@
+-module(fish).
+
+-compile(export_all).
+
+fish() -> fish.
--- a/inttest/thooks_exclude/rebar.config
+++ b/inttest/thooks_exclude/rebar.config
@@ -1,0 +1,7 @@
+%% pre-scripts
+{pre_hooks, [{clean, "echo preclean >> preclean.out"},
+             {compile, "echo precompile >> precompile.out"}]}.
+
+%% post-scripts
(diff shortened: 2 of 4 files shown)


tuncer:
Maybe this is better solved by rebar3's profile support. I'd like to see only one rebar version, given the limited development resources, and I hope we can eventually (once rebar3 is a smooth-to-migrate-to replacement) put rebar2 into critical-bugfixes-only mode.

derek121:
Agreed- that makes sense.

tsloughter:
Just to make expectations clear, "smooth-to-migrate" to rebar3 will always be relative/subjective. It will never be 1-to-1, especially for Makefiles or scripts that people rely on that run rebar commands that no longer exist.

PR #349, opened 2014-08-20
fixing detecting deps dir for cover.spec files

marcelog:
Hey guys,

I've hit a *possible* a bug when trying to run ct tests with deps that have a cover.spec file. It seems that collect_glob/3 does not correctly determine that it should skip a cover.spec file inside a dependency directory and it ends up seeing all available cover.spec files (the one for the root project and the ones inside their deps).

This patch seems to work ok. 

Any thoughts about it?

Thanks in advance!

### DIFF
--- a/inttest/deps_with_cover_spec_files/a.erl
+++ b/inttest/deps_with_cover_spec_files/a.erl
@@ -1,0 +1,6 @@
+-module(a).
+
+-compile(export_all).
+
+hello() ->
+    io:format("hello world\n").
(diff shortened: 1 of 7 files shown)


tuncer:
ping?

ferd:
The merge conflict is on 9713dafcb509eb39e9dddd8bba2d9591c77e7455:

> commit 9713dafcb509eb39e9dddd8bba2d9591c77e7455
> Author: YAMAMOTO Takashi <yamt@mwd.biglobe.ne.jp>
> Date:   Fri Aug 30 18:16:09 2013 +0900
>
>    Fix spec file look up
>
>    When trying to skip spec files under deps/ directory,
>   ignore "deps" component which is also included in Cwd.
>    For example, "/home/deps/src/myapp/test/cover.spec"
>    contains "deps" component but should not be skipped if
>    Cwd is "/home/deps/src/myapp/".

It looks like an old commit, but the line is not there in the diff for this PR and they appear to be fixing the same problem. I'm closing this, ping me if it turns out I was wrong.

PR #370, opened 2014-10-15
Fix CT suites crashing when deps have coverspec

zacharyfox:
This fixes a bug introduced in commit 5e91322e4ab0383068463844ead85ff15c0fba9e that causes rebar ct to fail when running where a dependency also has a cover.spec file.


tuncer:
ping?

zacharyfox:
This is a duplicate of #406. Closing without merge.

PR #395, opened 2014-11-21
Add colored logs

platinumthinker:


tuncer:
Ping?

As I said earlier, without the command-line switch (log_colored=), this is okay to merge, since color mode is default-off. Once that's in, we can still figure out how to expose overriding the compile-time default.

tuncer:
To get this landed, I went ahead and created a fixed version of the patch (#522). Once that lands, this can be closed as merged. Thanks, @platinumthinker.

PR #410, opened 2014-12-22
gpb_compile: restore missing gpb proto compile message

lrascao:
Conjunction of pull requests #386 and #408 (8f64e1, 81063d) led to compilation message being removed.

#408 removed the compilation message based on the fact that proto gpb compilation was making use of rebar's base compiler, #386 stopped making use of rebar's base compiler to overcome correct dependency checking issue.


tuncer:
Can this can be closed, now that #424 was submitted, or is it still relevant?

lrascao:
it can be closed since #424 fixes it

PR #412, opened 2014-12-23
Transform list of unicode chars in templater

studzien:
Since OTP 17 strings read from a file are by default parsed as a list of
unicode characters.
If such a list contains a non-ascii character (unicode > 255) then a
call to erlang:list_to_binary/1 fails.
This commit converts unicode list of chars to binary via unicode module
if ran against OTP >= 17.

### DIFF
--- a/inttest/rgen1/test.config
+++ b/inttest/rgen1/test.config
@@ -1,1 +1,2 @@
 {web_port, {{web_port}} }.
+{non_ascii_var, {{non_ascii_var}} }.
--- a/inttest/rgen1/vars.config
+++ b/inttest/rgen1/vars.config
@@ -1,1 +1,2 @@
 {web_port, 1234}.
+{non_ascii_var, "\"zażółć gęślą jaźń\""}.
--- a/src/rebar_templater.erl
+++ b/src/rebar_templater.erl
@@ -87,7 +87,13 @@ create(Config, _) ->
 resolve_variables([], Dict) ->
     Dict;
 resolve_variables([{Key, Value0} | Rest], Dict) when is_list(Value0) ->
-    Value = render(list_to_binary(Value0), Dict),
+    Value1 = case rebar_utils:otp_release() of
+        [$R|_] ->
+            list_to_binary(Value0);
+        _ ->
+            unicode:characters_to_binary(Value0)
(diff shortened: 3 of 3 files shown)


tuncer:
What's the status?

studzien:
I'm going to close this PR and swap the reltool vars interpreter from ``file:consult`` to something that simply reads the binary data and doesn't try to understand unicode characters (possibly ``rebar_templater:consult``). What do you think?

tuncer:
What's wrong with calling unicode:characters_to_binary/1 with OTP older than 17?

studzien:
Here's a short example that I guess shows the point of the issue.
Let's take a string consisting of a single letter, unicode 200 (```"È"```).
In binary UTF-8 file this string consists of two bytes: ``0xC3 0x88``. Prior to OTP 17, ``file:consult`` simply returns ``[195, 136]``, so calling ``unicode:characters_to_binary`` will return ``<<195, 130, 194, 187>>``, cause the function will interpret each of the list's entries as an unicode letter, what is obviously not true, since both letters combined form an UTF-8 letter.
On the other hand, this single letter will be read as ```[200]``` in OTP 17 and now, the function call returns desired ``<<195, 136>>``.
I think I'm starting to understand encoding now :)

tuncer:
ping?

tuncer:
ping?

ferd:
This is a month old since the last ping, and it's been over 4 months since the last update. I'Mm closing this. We can reopen it if new stuff happens.

studzien:
Hi.
Sorry, I wasn't really aware that I was the one pinged, after all my question above remains unanswered :)

Anyway, this is a really minor issue, happened to us once when we used a string with a Polish diacritic in a template overlay.

ferd:
Ah. Well.

> I'm going to close this PR and swap the reltool vars interpreter from file:consult to something that simply reads the binary data and doesn't try to understand unicode characters (possibly rebar_templater:consult). What do you think?

That sounds like it would make sense (in which case, this can remain closed)

studzien:
Sure, thanks.
Sorry for the misunderstanding, I should have reacted earlier, but somehow I missed the first ping :)

PR #417, opened 2014-12-29
Allow mib_opts to override snmpc options (fixes #416)

### LINKED ISSUE #416: Cannot specify a different include dir for snmpc using mib_opts in rebar.config
I have a scenario where I wish to compile a MIB in my project which depends on a MIB in a dependency. I set my rebar.config mib_opts to:
```
{mib_opts, [{i, ["deps/custom_snmp/priv/mibs"]}]}.
```
but snmpc fails to to find the the include dir. SNMPC only uses the first include directive which in the case of rebar is `{i, ["priv/mibs"]}`. I propose that mib_opts is given priority by putting them first in the list of options passed to `snmpc:compile`.

As a workaround I symlink the required MIBS into priv/mibs but now I am deploying the same MIB multiple times. Thoughts?

josephDunne:
make mib_opts have a higher priority over the rebar defaults. Allows the end use to override the include and output directory using mib_opts


tuncer:
Seems reasonable to me. @ferd?

tuncer:
ping?

ferd:
Relaunching the build to see if it still passes. Will merge if it does.

ferd:
All runs fail with this, not mergeable.

tuncer:
@josephDunne?

ferd:
It has been over 5 months without updates, and 3 weeks since the last ping. Closing this. We can reopen if anything new comes up.

PR #418, opened 2014-12-30
Fix #415 (reltool vsn check)

### LINKED ISSUE #415: Rebar generate not work if erlang compile manualy (Ubuntu)
I`am clone erlang from github and compile it. In my erlang folder in lib directory no version erlang library. Only simpe name (asn, common_test, ... reltool (not reltool-0.6.6) etc).  If i run 
```
rebar compile generate
```
i see error:
```
ERROR: Reltool support requires at least reltool-0.5.2; this VM is using reltool
```

Sorry for my english.

tuncer:
* Refactor reltool version check to use vsn app key if ebin/reltool.app
  exists.

  OTP releases prior to R14 didn't install the app file, so we still
  have to support getting the version string from the lib_dir path. If
  you're using an R13B03 or R13B04 install that doesn't have version
  strings in lib_dir paths (lib/reltool vs lib/reltool-0.5.3), we won't
  be able to retrieve the version string. Given R13's age and that it's
  unusual to have vsn-less lib_dir paths, this shouldn't be problem.

* While at it, delete trailing white space.

### DIFF
--- a/src/rebar_reltool.erl
+++ b/src/rebar_reltool.erl
@@ -117,7 +117,7 @@ check_vsn() ->
             ?ABORT("Reltool support requires the reltool application "
                    "to be installed!", []);
         Path ->
-            ReltoolVsn = filename:basename(Path),
+            ReltoolVsn = reltool_vsn(Path),
             case ReltoolVsn < "reltool-0.5.2" of
                 true ->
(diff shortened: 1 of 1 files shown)


ferd:
Works on local tests, merging.

PR #421, opened 2015-01-02
inttest/ct3: fix travis-ci breakage

tuncer:


tuncer:
Let's see if this works better on travis-ci.

tuncer:
Before:
```
{error_logger,{{2015,1,1},{13,48,39}},
    "Can't set long node name!\nPlease check your configuration\n",[]}
```

After:
```
{error_logger,{{2015,1,2},{21,22,11}},
  "Protocol: ~tp: the name ct_rt3@localhost seems to be in use by another Erlang node",
["inet_tcp"]}
```

ferd:
What happens if we generate a random name? I'm wondering if the issue is actually the name or something regarding their network and instances or something entirely different.

tuncer:
@BanzaiMan mentioned https://github.com/travis-ci/travis-cookbooks/commit/31b08f9db07f24604d160233a788c61a1b2c7ff1

tuncer:
Modified Makefile to test spinning up a distributed node in various ways.

ferd:
IP addresses belong to long names, not short names (at least for ipv4)

tuncer:
I've included it because it didn't fail locally. I'll disable it and push to see what else fails on travis-ci.
Note: `-sname ct3@::1` is okay, `-sname ct@127.0.0.1` is not, maybe EPMD doesn't support any IPv4.

tuncer:
Failed:
- `-sname ct3@127.0.0.1`
- `-name ct3`

Succeeded:
- `-sname ct3`
- `-name ct3@127.0.0.1`
- `-name ct3@::1`
- `-sname ct3@::1`

ferd:
So that's a sign about long names and short names. Erlang considers a domain to be long if it's got a period in it (`.`) and short otherwise. Or rather, it perceives a host with a `.` in it to not be short, and to accept mostly anything submitted for a long name:

https://github.com/erlang/otp/blob/maint/lib/kernel/src/net_kernel.erl#L1244-L1268

So `127.0.0.1` is forced long, but `::1` is either. I'm guessing that's why it fails to set the short name with a 127.0.0.1 address, while ct3@::1 works.

For the auto-detection, we'd have to look at the output of `{inet_db:gethostname(),inet_db:res_option(domain)}` which would explain fairly quickly why it fails.

tuncer:
Why does `-sname ct3@127.0.0.1` work locally? Does it work on your machine?

ferd:
It does not work on my machine.

ferd:
So you see the problem in the output: `{inet_db:gethostname(), inet_db:res_option(domain)} -> {"localhost",[]}` That domain value is not set. It can be set by hand with https://github.com/erlang/otp/blob/maint/lib/kernel/src/inet_db.erl#L211 , but in practice, it should be done by the system: https://github.com/erlang/otp/blob/172e812c491680fbb175f56f7604d4098cdc9de4/lib/kernel/src/inet_config.erl#L250-L261

https://github.com/erlang/otp/blob/172e812c491680fbb175f56f7604d4098cdc9de4/lib/kernel/src/inet_config.erl#L231-L245

I'm guessing this bit here has a single name found and ends up not setting both values. Not sure why that happens exactly.

tuncer:
Update: waiting for response from travis-ci regarding potential `/etc/hosts` correction.

tuncer:
Before I test the container-based workers suggested by @BanzaiMan, test binding to 127.0.0.1.

ferd:
Well that seems to work. If we can rewrite tests to use that, that would probably be best so our build status is accurate again. In any case I'm not sure we should depend on a properly configured host file when possible to avoid, given that may turn out contributors will also have a broken config on their local machine.

tuncer:
It's a last resort, and it has problems like being IPv4-specific and having to remember that any and all distributed nodes we start use `@127.0.0.1`.

tuncer:
Now, let's try container-based workers.

ferd:
That seemed to go well too.

tuncer:
Yes, @BanzaiMan's suggestion to use container-based workers fixed it.
I'd prefer to go with this, considering it worked correctly the way it's supposed to and the hostname looks sane (`ct_run -noshell [...] -name test@testing-worker-linux-docker-f95833d6-3233-linux-5.prod.travis-ci.org`).

tuncer:
Cleaned up branch and commit.

ferd:
Sweet, thanks for the fix. Merging.

Issue #384, opened 2014-11-03
gpb plugin check last modification not working correctly in specific case


lrascao:
when using PR #203 (@tuncer @tomas-abrahamsson @tsloughter @ferd) in my project that makes use of it (git@github.com:lrascao/erlang-srv-stub.git) i have an issue with the build using rebar. in this scenario i have a src/builtin.proto and src/builtin.erl, in rebar.config i define a module_name_suffix so that the generated file will not clash with the already present one, however this tricks rebar_base_compiler:run into thinking that there are no modifications (since both files are being tracked and have the same timestamps) so the build fails. repro steps summarized are:
- have x.proto and x.erl in src
- define a module name suffix in gpb_opts (eg. _pb)
- rebar compile won't build .proto file

lrascao:
fixed

PR #424, opened 2015-01-06
Gpb recompilation detection (using base compiler)

tomas-abrahamsson:
This is a second attempt at solving the #384, this time it uses the `rebar_base_compiler` for compiling. (The first attempt, was #386, which was reverted by #413)

The gpb supports both module name prefixes and suffixes, so the `rebar_base_compiler` has been extended to support this. So now in addition to the previous way of specifying source file name extension and target extension, it is now also possible to specify a `{Source,Target}` file names in full, as per the idea by @tuncer, in discussion also with @lrascao. The old way still works.

Now the proto compiler needs to find files by extension, so the `rebar_utils` has been extended with `find_files_by_ext/2,3`, which runs a regexp search to avoid OS X resource fork files. All relavent calls to `rebar_utils:find_files/2,3` have now been changed into `rebar_utils:find_files_by_ext/2,3` instead. In fact, the vast majority of the calls tried to find files by extension. One commit introduces `rebar_utils:find_files_by_ext/2,3`, and another commit changes all calls, as per suggestion by @tuncer. These two are the first commits on the branch.

An additional inttest has been added, by @lrascao to verify that nothing is recompiled when nothing has changed.

The commits on the branch have been discussed via mail with @tuncer and @lrascao prior to submission.

### DIFF
--- a/inttest/proto_gpb/proto_gpb_rt.erl
+++ b/inttest/proto_gpb/proto_gpb_rt.erl
@@ -29,6 +29,8 @@
          run/1]).
 
 -include_lib("eunit/include/eunit.hrl").
+-include_lib("kernel/include/file.hrl").
+-include_lib("deps/retest/include/retest.hrl").
 
 -define(MODULES,
         [foo,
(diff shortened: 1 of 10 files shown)


tomas-abrahamsson:
Fixed the silly xref build error, so now it is the "right" build error ("Can't set long node name! Please check your configuration")

tuncer:
cc @ferd

tuncer:
@tomas-abrahamsson, travis-ci issues will be resolved once you rebase against master.

tomas-abrahamsson:
Just rebased it.

tuncer:
cc @ferd @tsloughter

PR #426, opened 2015-01-10
Fixed #133. Release upgrade now handle long and short names properly.

### LINKED ISSUE #133: upgrade script does not handle long node name properly
Node runner script does not respect name type in vm.args and runs ```install_upgrade.escript``` which itself only works with shortnames, and finally tries to start node with name like ```node@127.0.0.1_upgrader_18328``` and of course fails.

seriyps:
Implemented, by moving upgrade installation from install_upgrade.escript to nodetool, which already supports both short and long names.

See https://github.com/rebar/rebar/issues/133#issuecomment-69211063 for details.

I've tested this version on my project and it works nice. No tests (`make test`) were broken.

### DIFF
--- a/priv/templates/simplenode.reltool.config
+++ b/priv/templates/simplenode.reltool.config
@@ -38,6 +38,7 @@
            {copy, "files/{{nodeid}}", "bin/{{nodeid}}"},
            {copy, "files/{{nodeid}}.cmd", "bin/{{nodeid}}.cmd"},
            {copy, "files/start_erl.cmd", "bin/start_erl.cmd"},
+           %% Following line may be safely removed in new projects
            {copy, "files/install_upgrade.escript", "bin/install_upgrade.escript"},
            {copy, "files/sys.config", "releases/\{\{rel_vsn\}\}/sys.config"},
            {copy, "files/vm.args", "releases/\{\{rel_vsn\}\}/vm.args"}
--- a/priv/templates/simplenode.install_upgrade.escript
+++ b/priv/templates/simplenode.install_upgrade.escript
@@ -2,51 +2,54 @@
 %%! -noshell -noinput
 %% -*- mode: erlang;erlang-indent-level: 4;indent-tabs-mode: nil -*-
 %% ex: ft=erlang ts=4 sw=4 et
+%% This file is left for backward-compatibility.
+%% You, probably, shouldn't include it to new projects.
 
--define(TIMEOUT, 60000).
--define(INFO(Fmt,Args), io:format(Fmt,Args)).
-
-%% TODO: This script currently does NOT support slim releases.
-%% Necessary steps to upgrade a slim release are as follows:
-%% 1. unpack relup archive manually
-%% 2. copy releases directory and necessary libraries
-%% 3. using release_hander:set_unpacked/2 .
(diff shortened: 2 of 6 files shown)


tuncer:
+1

ferd:
Just to be clear, that's a backwards incompatible change over the current tooling we provide, isn't it?

tuncer:
Yes, as the functionality is moved into a sub-command of nodetool. Do you think `install_upgrade.escript` should be modified to call `nodetool upgrade` with maybe a deprecation message?

ferd:
Either that or just left there for compatibility reasons. Otherwise we'll have to go a full deprecation cycle on this PR.

seriyps:
So, should I add some wrapper to this PR, as @tuncer suggests?

ferd:
Yeah. That way we'd maintain backwards compatibility as much as possible. Is there anything else you'd expect to break or change for someone using the script?

seriyps:
No, I'll just add a wrapper for nodetool, which will follow original `install_upgrade.escript` API and will show warning message and then delegate all the work to nodetool via, say, `os:cmd` or `open_port`. Nothing else should be broken.

ferd:
Then that sounds good.

seriyps:
@ferd maybe I should add comment about deprecation to `{copy, "files/install_upgrade.escript", "bin/install_upgrade.escript"}` line in default `reltool.config`?

ferd:
Yeah. That one would be needed. In practice I don't know if this will be considered breaking or not, if existing configs suddenly stop working because we don't copy the file `install_upgrade.escript` needs to work under the new code with an old config.

It would be extremely annoying to deploy a node/release and then when you need to operate on it, suddenly realize the scripts don't work because someone upgraded rebar on the build server.

seriyps:
@ferd ok, I've just added backward-compatible `install_upgrade.escript` as a wrapper for `nodetool`.

ferd:
Alright, with the rebasing stuff looks like it still passes. I'm in for the backwards compatible script. @tuncer any last thing before I merge this in?

priv/templates/simplenode.reltool.config
```diff
@@ -38,7 +38,7 @@
            {copy, "files/{{nodeid}}", "bin/{{nodeid}}"},
            {copy, "files/{{nodeid}}.cmd", "bin/{{nodeid}}.cmd"},
            {copy, "files/start_erl.cmd", "bin/start_erl.cmd"},
-           {copy, "files/install_upgrade.escript", "bin/install_upgrade.escript"},
+           {copy, "files/install_upgrade.escript", "bin/install_upgrade.escript"},  % may be safely removed in new projects
```
tuncer: I would put the comment on a separate (previous) line for readability.

seriyps: @tuncer fixed in ab4c20f

tuncer: Thanks, the new version has to use `%%` though, and it should be indented like the surrounding lines.

tuncer: Also, it's best to squash the fixup commits into the original commit.

seriyps:
@tuncer comment fixed and all commits squashed to one.

tuncer:
@seriyps okay

PR #428, opened 2015-01-12
Remove a git url in gpb and protobuffs inttest configs

tomas-abrahamsson:
In the rebar.config files for the gpb and protobuffs inttests, use the
`{Name, Version}` dependency form instead of `{Name, Version, Source}`,
since the `Source` is not really applicable; the mocks are copied into
place by the proto_gpb_rt:files/0 and proto_protobuffs_rt:files/0.
Idea and suggestion by @tuncer.


tuncer:
+1

ferd:
Worked on local tests, +1.

PR #430, opened 2015-01-13
Exit with proper status code if 'eval' fails

robertoaloi:


tuncer:
Seems reasonable. Would it make sense to extract this into a helper function in the script?

tuncer:
Seems reasonable. Would it make sense to extract this into a helper function in and update the whole script?

ferd:
This change makes sense. Where else would you see it, @tuncer ?

I believe this also needs to be rebased on master so tests pass.

tuncer:
I was suggesting that we might use the opportunity to make it a shell function and reuse everywhere $? is checked.

robertoaloi:
@ferd Rebased on latest master.
@tuncer I agree some refactoring would be nice for the script, but I would consider that a separate issue from this one.

tuncer:
> but I would consider that a separate issue from this one

Works for me.

PR #434, opened 2015-01-17
Make rebar xref recognize -behavior()

paulo-ferraz-oliveira:
Erlang allows for -behavior as well as -behaviour (even rebar uses a mix of these) so the check should ignore the callbacks in both cases.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -133,3 +133,4 @@ Vlad Dumitrescu
 stwind
 Pavel Baturko
 Igor Savchuk
+Paulo F. Oliveira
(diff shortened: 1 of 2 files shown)


paulo-ferraz-oliveira:
I will correct the commit message... Just now figured it goes beyond 50 chars per line

ferd:
Am I right in guessing this is related to https://github.com/rebar/rebar/issues/435 ? 

The change otherwise looks sane to me.

paulo-ferraz-oliveira:
The issues are not related except for the fact that I came across them at the same time, but I'd never used rebar xref before... the other issue is related to the fact that xref warnings are output as errors (even to stderr), which stops an ongoing build process (e.g., make)

tuncer:
What's the status?

ferd:
I forgot this PR. I'll go merge it. I figure the conflict will be around the THANKS file.

PR #439, opened 2015-01-21
Set +A0 among emulator args for rebar's escript

robertoaloi:
While switching from R15 to R16, we noticed a significant performance degradation when compiling our code with rebar (compilation time almost doubled).
Adding the "+A0" flag seems to have an extremely positive effect, reverting compile times to the R15 values.
This seems related to rebar performing lot of IO from a single process.
Not sure if the patch is general enough to be part of the upstream rebar, but I decided to open this PR to get the discussion started.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -4,6 +4,7 @@
 %% escript_incl_extra is for internal rebar-private use only.
 %% Do not use outside rebar. Config interface is not stable.
 {escript_incl_extra, [{"priv/templates/*", "."}]}.
+{escript_emu_args, "%%! -pa rebar/rebar/ebin +A0\n"}.
 
 %% Types dict:dict() and digraph:digraph() have been introduced in Erlang 17.
 %% At the same time, their counterparts dict() and digraph() are to be


tuncer:
Have you compared R15 `+A10` vs `+A0` (default)?

robertoaloi:
Will try now.

robertoaloi:
@tuncer From what I can see, in R15 the default is +A0, so my patch should not have any effect. Starting from R16 the default is +A10, so my patch revert the old behaviour for the rebar escript and in that case we noticed the huge performance increase. Right?

tuncer:
Yes, and it's a good idea to first compare R15 in both settings, to hopefully rule out other changes included in R16.

ferd:
I've seen this before and know what this is about yes. I'm fine with going back to R15 settings (especially since rebar worked fine there). The cost of setting up async threads isn't really an advantage for an ideally short-running app like Rebar.

tuncer:
I've read @robertoaloi's post to mean it's a slowdown during compilation, and not just setup overhead, but I may be wrong. @ferd, when you say "I've seen this before", do you mean just the increase in startup time?

ferd:
Yeah. This set of flags impacts every time the VM is booting by starting N async OS threads rather than just letting the VM handle it all. We don't really need them in a sequential code base like rebar for the most part.

tuncer:
@psyeugenic has at least twice advised us against using `+A0`, and doing that may even have a worse effect beginning with R16, as `+A10` is the default now. However, we've successfully used the default (`+A0`) in rebar from R13 up to R15B03-1.

In either case it's important to measure what's happening first. So, as suggested profile R15 with +A0 and +A10, and then the same for R16 (plus maybe R17).

With all that said, I'm not strongly against `+A0`, but at the very least compilation is not sequential as we spin up worker processes for that. So, rebar isn't a single sequential process.

psyeugenic:
I'm not surprised that it faster with +A0 since file commands are not scheduled on the async threads.

You could be fine using file commands in the scheduler threads but I doubt you could count on it. It is especially problematic on slow discs or over NFS. One possible symptom is scheduler collapse, same as for long running NIFs.

richcarl:
Note that the erlc command is using +A0, so it's obviously not that dangerous.
With regards to parallelism of rebar - the tradeoff when using asynch threads for IO is latency vs throughput. When you're running a task like rebar, the main thing that the user cares about is when the entire job is done, hence throughput is the main concern. When running something like a web server, it's more important that each connecting client gets a response soon, even though the total transfer time will suffer when there's only a single client connecting. The latter is the typical mode of an Erlang program, which is why asynch threads is now the default. But it doesn't make much sense for a command line tool - in that case, it's better to let one process wait while another is reading a file, if the total throughput gets higher.

ferd:
@psyeugenic my understanding is really just that starting 10 threads is slower than starting 0, and that if the files we look for and their operations aren't scheduled on these 10 threads, we'd get a speedup at start time (which is a huge part of our total run time in rebar) by disabling these.

Personally my opinion falls in line with Richard's. I'm going to leave this open for a while for more comments but right now I'm of the opinion we should merge this in.

tuncer:
As I said before, I'm not strongly against +A0. rebar has been used that way since the beginning, and assuming `+A0` in 17.x and later isn't worse than R16 and earlier, I can see this get merged.

Also, @robertoaloi hasn't answered yet, but I'll take @richcarl's response to mean that @robertoaloi was talking about startup slowdown and not at all about overall compilation slowdown. @richcarl, is that correct?

@psyeugenic, @richcarl, I'm not saying we should use it, but out of curiosity, what's your opinion on `+A1`? Any measurements?

tuncer:
As I said before, I'm not strongly against `+A0`, and rebar has been used that way since the beginning, so I can see this get merged.

Also, @robertoaloi hasn't answered yet, but I'll take @richcarl's response to mean that @robertoaloi was talking about startup slowdown and not at all about overall compilation slowdown. @richcarl, is that correct?

@psyeugenic, @richcarl, I'm not saying we should use it, but out of curiosity, what's your opinion on `+A1`? Any measurements?

psyeugenic:
@richcarl All I'm saying is that you open up for undefined behaviors .. it is however unlikely it will happen, it all depends.
@ferd Is it the start time that's killing you? I'm pretty sure it's the latency of handing over file-ops between threads. I haven't measured it though.
@tuncer No measurement. Turn it of or stick with +A10 (or how many parallel file ops you have) I think many Ericsson projects use +A256.

I can reiterate the problem point many times and you know the pros and cons .. now make a decision. =)

richcarl:
@psyeugenic Sure, but does what you're saying also imply that OTP are worried about the use of +A0 in erlc, and have plans to stop using it (at a cost of possibly much slower compilation)?

psyeugenic:
@richcarl Well, don't do as we do, be better! =) I was actually under the impression that erlc used non-smp which would have made more sense with +A0 but nooo. We will probably change erlc to +A10 (or whatever default we find reasonable) if we have more then one file to compile but use non-smp with +A0 in the single file case.

We reason that in the non-smp case only the soft-realtime properties are affected. Since the compiling is more or less sequentially handled it's therefor a none-issue in our case.

My point remain, if you want to be safe use async-threads.

tuncer:
I wish we could spin up +A threads from within a running VM.

rebar.config
```diff
@@ -4,6 +4,7 @@
 %% escript_incl_extra is for internal rebar-private use only.
 %% Do not use outside rebar. Config interface is not stable.
 {escript_incl_extra, [{"priv/templates/*", "."}]}.
+{escript_emu_args, "%%! -pa rebar/rebar/ebin +A0\n"}.
```
tuncer: Can you insert a blank line or alternatively add a comment explaining why we set `+A0`?

tuncer:
@psyeugenic wrote:
> We will probably change erlc to +A10 (or whatever default we find reasonable) if we have more then one file to compile but use non-smp with +A0 in the single file case.

That's easy to do via erlc's C wrapper, and I wish we could spin up extra async-threads from within a running VM.

At least we can always override `emu_args` via `ERL_FLAGS=+An rebar`.

tuncer:
The question of whether it's just startup time or overall compilation time is still unanswered, and it'd also be great to know if `+A1` is a good compromise, but I'd like to get this resolved.
So, perhaps we should make it `+A0` and re-evaluate when the mentioned erlc changes have been made or we find out `+A1` is a good choice.

tsloughter:
Rebar3 has been using +A0  for a while now.

I'm also considering `-smp disable` I can't remember now why I hadn't already.

tuncer:
If you disable SMP, there's not much point in keeping compiler workers.

tuncer:
Also, with things like Dialyzer plugins, it's even more important to have SMP, except if you're memory limited and have to enable +S1 to avoid a Dialyzer OOM condition.

tuncer:
@robertoaloi, can you measure what difference these make in your env compared to an unmodified rebar with R16B03 (and 17.4.1):

1. `-mode minimal`
2. `-mode minimal +A1`
3. `-mode minimal +A2`

For interesting results, try these with R13B04 :).

Ultimately, there's more than one slow down since the R13 days, so OTP (starting with erts) is in need of profiling and perf fixing. I believe we can get it mostly back, but I haven't tried yet.

tsloughter:
Yes, we aren't disabling smp in rebar3 because of dialyzer and ct.

tuncer:
@robertoaloi, @richcarl, I'm afraid I have to repeat the question: is it just startup overhead or overall compile times? I don't know how exactly you measured, but a simple timing here doesn't result in more than 30ms reduction with +A0 vs +A1 or +A20.

ferd:
I believe it's usually startup time. We can close this if you measured near nil improvements. On what platform did you measure it, though? Some are worse than others for start times.

tuncer:
Timing normal (release) rebar binary on 18.0-rc1 ce96ab6:

Flags | `time rebar version` real (ms)
------------ | -------------
`+A0` | 280
`+A0 -mode minimal` | 240
`+A0 -boot start_clean` | 280
`+A0 -boot start_clean -mode minimal` | 240
`+A1` | 310
`+A10` | 310
`+A10 -mode minimal` | 280
`+A10 -boot start_clean` | 310
`+A10 -boot start_clean -mode minimal` | 280
`+A20` | 310

This is still comfortably below 400ms and therefore fast enough to not disrupt the flow, but barely so.

It seems impossible to enable the initially disabled bits like distrib once the VM has been started in `-mode minimal` mode. Is there really no way?

For contrast, here's R13B03 after rebuilding rebar with R13B03:

Flags | `time rebar version` real (ms)
------------ | -------------
`+A0` | 160
`+A0 -mode minimal` | 130
`+A0 -boot start_clean` | 130
`+A0 -mode minimal -boot start_clean` | 130
`+A10` | 190
`+A10 -mode minimal` | 150
`+A10 -boot start_clean` | 190
`+A10 -mode minimal -boot start_clean` | 150

tuncer:
Measurement took place on a modern linux-amd64 distro with an otherwise idle machine.

tuncer:
Some results from a modern linux-amd64 distro.

Timing normal (release) rebar binary on 18.0-rc1 ce96ab6:

Flags | `time rebar version` real (ms)
------------ | -------------
`+A0` | 280
`+A0 -mode minimal` | 240
`+A0 -boot start_clean` | 280
`+A0 -boot start_clean -mode minimal` | 240
`+A1` | 310
`+A10` | 310
`+A10 -mode minimal` | 280
`+A10 -boot start_clean` | 310
`+A10 -boot start_clean -mode minimal` | 280
`+A20` | 310

This is still comfortably below 400ms and therefore fast enough to not disrupt the flow, but barely so.

It seems impossible to enable the initially disabled bits like distrib once the VM has been started in `-mode minimal` mode. Or is there a way?

For contrast, here's R13B03 after rebuilding rebar with R13B03:

Flags | `time rebar version` real (ms)
------------ | -------------
`+A0` | 160
`+A0 -mode minimal` | 130
`+A0 -boot start_clean` | 130
`+A0 -mode minimal -boot start_clean` | 130
`+A10` | 190
`+A10 -mode minimal` | 150
`+A10 -boot start_clean` | 190
`+A10 -mode minimal -boot start_clean` | 150

ferd:
18-RC will be particularly slow because they use one timer wheel per scheduler afaict, and they still need to optimize that. Here's the same for me on 17.3 on OSX 10.9.5:

Flags | `time rebar version` total (ms)
------------ | -------------
nothing | 387
`+A0` | 389
`+A0 -mode minimal` | 388
`+A0 -boot start_clean` | 387
`+A0 -boot start_clean -mode minimal` | 382
`+A1` | 388
`+A10` | 385
`+A10 -mode minimal` | 401
`+A10 -boot start_clean` | 385
`+A10 -boot start_clean -mode minimal` | 388
`+A20` | 373

There's no visible improvement for me here, everything more or less fits between a 30ms jitter window. At least for rebar boot time itself I can't say I see any difference here.

ferd:
E: I had a version where I badly measured effects. Here's the proper re-run.

18-RC will be particularly slow because they use one timer wheel per scheduler afaict, and they still need to optimize that. Here's the same for me on 17.3 on OSX 10.9.5:

Flags | `time ./rebar version` total (ms)
------------ | -------------
nothing | 377
`+A0` | 306
`+A0 -mode minimal` | 270
`+A0 -boot start_clean` | 309
`+A0 -boot start_clean -mode minimal` | 277
`+A1` | 371
`+A10` | 442
`+A10 -mode minimal` | 314
`+A10 -boot start_clean` | 385
`+A10 -boot start_clean -mode minimal` | 371
`+A20` | 387

There's some improvement for me here, I also observed a 30ms jitter window, or ±15ms. It does seem to change up to about 100ms (or 25%) optimistically.

tuncer:
@ferd, are you sure there aren't outliers in your listing? Specifically, is `+A10` always >400, and what about `+A10 -boot start_clean -mode minimal`?

ferd:
10 runs of `+A10`: 763, 418, 387, 444, 404, 392, 384, 401, 407, 397

10 runs of `+A10 -boot start_clean -mode minimal`: 325, 341, 487, 345, 337, 335, 335, 346, 349, 348

There's gonna be lots of outliers on any listings is my guess.

psyeugenic:
The startup slowness in 18-rc1 was mostly due to erroneous large timer wheels. I did some measurement on 18 as well using `time erl -noshell -run erlang halt`:

| Version | real 
|------------|------
|`OTP-18.0-rc1-520-gcc03bb8`|105ms        
|`OTP-18.0-rc1-503-gce96ab6`|124ms
|`OTP-17.4`|107ms

Seems to be better.
We haven't looked at optimizing startup time all that much. It's not in the *need* category but more in the *nice to have* category.

tuncer:
Version | `time erl -noshell -run erlang halt` real
---- | ----
R13B03 | 137ms
OTP-18.0-rc1-503-gce96ab6 | 286ms

Version | `time erl +A0 -noshell -run erlang halt` real
---- | ----
R13B03 | 139ms
OTP-18.0-rc1-503-gce96ab6 | 258ms

Version | `time erl -mode minimal -noshell -run erlang halt` real
---- | ----
R13B03 | 102ms
OTP-18.0-rc1-503-gce96ab6 | 221ms

psyeugenic:
Yeah we already know that `OTP-18.0-rc1-503-gce96ab6` has worse characteristics. Measure against `OTP-18.0-rc1-520-gcc03bb8` instead .. or wait for rc2 next week.

tuncer:
erlang/otp.git is at rc1-503, and that's the newest I have, so how do I fetch the -520 branch? Am I supposed to get Rickard's OTP-12650?

psyeugenic:
Oh it isn't pushed  .. use my repo https://github.com/psyeugenic/otp that's synced. Yeah it's Rickards time merge at the tip.

tuncer:
Here's the results with a fresh build of OTP-18.0-rc1-520-gcc03bb8. Thanks whoever was responsible (@rickard-green?). The gap to R13B03 is still wide, but it looks considerably better. Another 50ms and we're getting there :).

Cmd | real
---- | ----
time erl -noshell -run erlang halt | 220ms
time erl -mode minimal -noshell -run erlang halt | 160ms
time erl +A0 -noshell -run erlang halt | 200ms
time erl +A0 -mode minimal -noshell -run erlang halt | 150ms
time ERL_FLAGS="+A0" rebar version | 217ms
time ERL_FLAGS="+A0 -mode minimal" rebar version | 178ms
time ERL_FLAGS="+A0 -mode minimal -boot start_clean" rebar version | 178ms
time rebar version | 240ms
time ERL_FLAGS="-mode minimal" rebar version | 200ms
time ERL_FLAGS="+A20" rebar version | 240ms
time ERL_FLAGS="+A20 -mode minimal" rebar version | 200ms

Time `rebar list-templates` inside rebar repo, after building it first:

```
$ cd rebar
$ make distclean all
```

Cmd | real
---- | ----
time ./rebar l-t | 522ms
time ERL_FLAGS="+A0" ./rebar l-t | 409ms
time ERL_FLAGS="-mode minimal" ./rebar l-t | 490ms
time ERL_FLAGS="+A0 -mode minimal" ./rebar l-t | 365ms

tuncer:
@ferd, for comparison, can you rerun with @psyeugenic's master branch?

ferd:
No, not for a couple of days at least. I'd need to set up an environment proper to this and today's Mothers' day here. Does it matter with regard to the issue at hand?

tuncer:
Not really, I just wanted to see how it affects OS X.

tuncer:
I think it's safe to say this can be closed.

PR #295, opened 2014-06-11
Travis otp vsns

tuncer:
Try out some otp version changes in .travis.yml.

PR #443, opened 2015-01-23
.travis.yml: extend otp vsn coverage

tuncer:
@ferd suggested we should test against all OTP versions supported by rebar


tuncer:
Closing, because I've added all otp versions in #295.

PR #444, opened 2015-01-25
Fix incorrect "not an app dir" warning

tuncer:


src/rebar_deps.erl
```diff
@@ -457,8 +457,19 @@ is_app_available(Config, App, VsnRegex, Path, _IsRaw = false) ->
                               {AppFile, {expected, App}, {has, OtherApp}}}}}
             end;
         false ->
-            ?WARN("Expected ~s to be an app dir (containing ebin/*.app), "
-                  "but no .app found.\n", [Path]),
+            case filelib:is_dir(Path) of
+                true ->
+                    %% Path is a directory, but it's not an app dir.
+                    ?WARN("Directory expected to be an app dir, "
+                          "but no .app found:~n~s~n", [Path]);
```
ferd: This error message makes it sound like the `.app` file should be at the root of the app dir, which is false. Sounds like keeping a mention of `ebin/*.app` would be useful.

tuncer: I changed it because we also look for `src/*.app.src[.script]`. Any suggestions for an improved message?

ferd: Maybe?
```erlang
?WARN("Directory expected to be an app dir, "
      "but no app file found in ebin/ or src/:~n~s~n", [Path]);
```

tuncer: Thanks, updated.

Issue #429, opened 2015-01-12
rebar doesn't respect the order of erl_first_files given in the rebar.conf file


norton:
The problem is that the sorted order of AllErlFiles does not match with the order of ErlFirstFilesConf. The order of ErlFirstFilesConf is specified by the user. The order of AllErlFiles determines the order of ErlFirstFiles.

Please review this snippet of code (https://github.com/rebar/rebar/blob/master/src/rebar_erlc_compiler.erl#L309-L313) to check my reasoning.  Git blame seems to point to this change (rebar/rebar@49c2564).

I have a rough idea for the code changes to address this regression.  However, I would like to confirm if there is a specification for the processing order of the erl_first_files given in the rebar.conf file.  My expectation is the files should be compiled in the configuration specified order.

nevar:
Hi.
While fixing problem with "compile non exist file" break "compile order" =(
I will look at this issue.

norton:
Great! Thank you.

tuncer:
Thanks, @nevar.

nevar:
@norton you can try https://github.com/nevar/rebar/tree/erlc_order for fix.
I need add test before do PR.

norton:
Yes, it works fine.  Thank you.

```
$ ./rebar --version
rebar 2.5.1 R15B03 20150116_040132 git 2.5.1-95-g4eef894

$ make
./rebar compile
==> qc (compile)
Compiled src/qc_gen.erl
Compiled src/qc.erl
Compiled src/qc_gen_http.erl
Compiled src/qc_statem.erl
Compiled src/qc_slave.erl
==> pmod_transform (compile)
Compiled src/pmod_pt.erl
==> ubf (compile)
Compiled src/contract_lex.xrl
Compiled src/contract_yecc.yrl
Compiled src/contract_lex.erl
Compiled src/contract_yecc.erl
Compiled src/ubf_utils.erl
Compiled src/contract_parser.erl
Compiled src/ubf_types_builtin.erl
Compiled src/contracts.erl
Compiled src/contract_driver.erl
Compiled src/contract_manager.erl
Compiled src/contract_proto.erl
Compiled src/proc_socket_server.erl
Compiled src/ubf_plugin_handler.erl
Compiled src/ebf.erl
Compiled src/contracts_abnf.erl
Compiled src/contract_manager_tlog.erl
Compiled src/ebf_driver.erl
Compiled src/qc_ubf_impl.erl
Compiled src/proc_utils.erl
[...]

PR #445, opened 2015-01-26
rebar doesn't respect the order of erl_first_files given in the rebar.conf file

norton:
The first commit is a fix for this issue.  The second commit is a test for this issue.  The second commit will fail if cherry-picked to the current master branch.

### DIFF
--- a/inttest/erlc/erlc_rt.erl
+++ b/inttest/erlc/erlc_rt.erl
@@ -31,7 +31,8 @@
 -include_lib("eunit/include/eunit.hrl").
 
 -define(MODULES,
-        [first_xrl,
+        [after_first_erl,
+         first_xrl,
          first_yrl,
          first_erl,
          foo,
@@ -42,7 +43,8 @@
          'SIMPLE-ASN']).
 
 -define(BEAM_FILES,
-        ["first_xrl.beam",
+        ["after_first_erl.beam",
+         "first_xrl.beam",
          "first_yrl.beam",
          "first_erl.beam",
          "foo.beam",
--- a/inttest/erlc/extra-src/after_first_erl.erl
+++ b/inttest/erlc/extra-src/after_first_erl.erl
@@ -1,13 +1,11 @@
 %% -*- mode: erlang;erlang-indent-level: 4;indent-tabs-mode: nil -*-
 %% ex: ts=4 sw=4 ft=erlang et
--module(first_erl).
+-module(after_first_erl).
+-compile({parse_transform, first_erl}).
 
 -include_lib("eunit/include/eunit.hrl").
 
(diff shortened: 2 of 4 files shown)


norton:
This merge request is for issue #429.

ferd:
So just so I'm clear, was the issue that `lists:partition` does not keep the expected order of files?

norton:
No.  The root issue is that AllErlFiles is not sorted to respect the order of ErlFirstFiles.

tuncer:
> The root issue is that AllErlFiles is not sorted to respect the order of ErlFirstFiles.

So, should we sort in `gather_src/2`, then?

norton:
I don't think so.  The fix made by Slava is fine by me.

nevar:
After we make `lists:partition` we loose order in `ErlFirstFilesConf`. We use order from file system and that is where bug came. We loose dependency order from user defined order.

ferd:
Ok, that makes sense to me. @tuncer anything else you can think of before I merge this in later today?

tuncer:
We can merge this, and I had a quick chat with @nevar that may or may not result in a follow-up ticket.

tuncer:
We can merge it as a fix, even though I don't like the extra list traversals too much.

norton:
Just checking ... Is anything else required for this pull request?

nevar:
@norton If you complete remove erl_first_files from config does rebar compile file in right order?

My main issue with this that erl_first_files use as is and don't order by rebar inner [algorithm](https://github.com/rebar/rebar/blob/master/src/rebar_erlc_compiler.erl#L320). So we need remove don't exists file and hope that user use correct order for this file.

I think that better ignore erl_first_files and use rebar algorithm for order files. But I not sure that all use case didn't use erl_first_files for some interesting behavior or rebar correct parse and analyze all posible erl file.

We can deprecate erl_first_files and lay on rebar order algorithm and improve it. Or keep erl_first_files and have more and more code for that options.

So my main question: Why we need erl_first_files if rebar can order erl file by himself?

norton:
@nevar fixing the behavior of erl_first_files and deprecating el_first_files are different goals.  My goal for this issue is simply to fix the erl_first_files regression reported in #429 and not introduce any new harm.

tuncer:
I've been thinking of ways to avoid the extra scanning, and didn't come up with a good solution. In order to fix the issue, we should get this in and try to optimize it in another patch.

tuncer:
To deprecate or remove `erl_first_files`, we'd have to be sure there are no valid cases where `erl_first_files` is used for something not covered by dependency resolution.

norton:
+1

norton:
+1

PR #459, opened 2015-02-26
Remove -m64 flag.

cmeiklejohn:
On 64-bit Linux, remove the emulation mode flag, given -m64 is an
invalid emulation mode on 64-bit Linux.

Example compilation from Basho's erlang_js on Linux:

```
ld: unrecognised emulation mode: 64
Supported emulations: elf_x86_64 elf32_x86_64 elf_i386 i386linux elf_l1om elf_k1om i386pep i386pe
make[4]: *** [Linux_All_OPT.OBJ/libjs.so] Error 1
```

### DIFF
--- a/src/rebar_port_compiler.erl
+++ b/src/rebar_port_compiler.erl
@@ -607,7 +607,7 @@ default_env() ->
      %% Linux specific flags for multiarch
      {"linux.*-64$", "CFLAGS", "-m64 $CFLAGS"},
      {"linux.*-64$", "CXXFLAGS", "-m64 $CXXFLAGS"},
-     {"linux.*-64$", "LDFLAGS", "-m64 $LDFLAGS"},
+     {"linux.*-64$", "LDFLAGS", "$LDFLAGS"},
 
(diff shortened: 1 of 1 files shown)


zeeshanlakhani:
:+1: 3bc4375

cmeiklejohn:
This PR would revert changes made by the following commit: https://github.com/rebar/rebar/commit/6fca763c9bb70d370d6782c436d5fca286db77df, which attempted to correct build issues on a 64-bit kernel and packaging running in a 32-bit userspace.

Vagabond:
I think this is correct, the linker doesn't usually accept arguments that specify architecture.

tsloughter:
And this is why rebar3 doesn't have a C code compile task... :)

aberghage:
+1 — architecture flags are valid for cc, not ld. If the compiler frontend does the linking, it might strip it off (IIRC GCC does), but generally we shouldn't be passing these to the linker.

cmeiklejohn:
Thanks!

PR #460, opened 2015-03-01
Add Dialyzer plugin

tuncer:
Build project-specific PLT
`$ rebar build-plt`

Check the PLT for consistency and rebuild it if it is not up-to-date
`$ rebar check-plt`

Analyze the code for discrepancies
`$ rebar dialyze`

Delete project-specific PLT
`$ rebar delete-plt`

Valid rebar.config options:

```
{dialyzer,
 [
  %% Store PLT in ~/.rebar/plt (Default)
  {plt_location, shared},
  %% Store PLT locally inside the project in .rebar
  {plt_location, local},
  %% Store PLT in custom directory
  {plt_location, "custom_dir"},
  %% Extra apps to include in the PLT
  {plt_extra_apps, [app1, app2]},
  {warnings, [unmatched_returns, error_handling]}
 ]}.
```


ferd:
So Dialyzer was previously removed from rebar. What makes you wish to bring it back?

tuncer:
1. rebar's knowledge of apps to include in the PLT and that motivating people to have a correct `applications` field in their app file
2. a common way of having project-specific PLTs
3. a simple solution to ignoring specific warnings and reporting only what's deemed relevant
4. Dialyzer education and use has changed since then

To be clear, this isn't based on the old code.

lrascao:
:+1:

tuncer:
We need a better naming scheme (maybe inside `~/.rebar`) for the project specific PLTs to avoid naming clashes. Any suggestions?

ebin/rebar.app
```diff
@@ -52,7 +53,11 @@
                   compiler,
                   crypto,
                   syntax_tools,
-                  tools]},
+                  tools,
+                  diameter,
```
ferd: Again I forgot. Why do we depend on this? Did we have diameter compiler support or something?

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,204 @@
...
+
+%% ===================================================================
+%% Public API
+%% ===================================================================
+
+%% TODO:
+%%
+%%  * How can we configure the filtering of warnings without exact line
+%%    numbers as a way to avoid updating the config whenever the file
+%%    in question is modified (and a warning's line number changes)?
```
ferd: Regular expressions?

ebin/rebar.app
tuncer: Yes, `rebar_dia_compiler`.

src/rebar_dialyzer.erl
tuncer: I was thinking of using `_` in the `dialyzer_ignored_warnings` list and subtracting that from a preprocessed `dialyzer:run/1` result. Not sure how regexes would apply.

ebin/rebar.app
ferd: This app didn't exist in older copies of Erlang and this may break stuff?

tuncer: It didn't exist, but what can break?

We could add it via a `dialyzer_extra_app` list in rebar.config as an alternative, to keep the .app file portable.

src/rebar_dialyzer.erl
tuncer: I was thinking of using `'_'` in the `dialyzer_ignored_warnings` list and subtracting that efficiently from a preprocessed `dialyzer:run/1` result. Can you think of a more efficient strategy, and do we need anything more specific than `'_'`, which means "any line", like, say, "lines x to y"?

tuncer: I was thinking of using `'_'` in a separate `dialyzer_ignored_warnings_noline` (name to be determined) list and subtracting that efficiently from a preprocessed `dialyzer:run/1` result. Can you think of a more efficient strategy, and do we need anything more specific than `'_'`, which means "any line", like, say, "lines x to y"?

tuncer: I was thinking of using `'_'` in `dialyzer_ignored_warnings` or a separate `dialyzer_ignored_warnings_noline` (name to be determined) list and subtracting that efficiently from a preprocessed `dialyzer:run/1` result. Can you think of a more efficient strategy, and do we need anything more specific than `'_'`, which means "any line", like, say, "lines x to y"?

tuncer:
I didn't add tests because running them would take too long.

tuncer:
Updated to store the PLT inside `~/.rebar/plt`.

ferd:
I'm all for including this, though I haven't had the time to try it yet. Might want to including in the next point release, which I'm really freaking late on cutting.

tuncer:
Yeah, let's include it in the next release.

ferd:
I'll take note to try it on some work projects tomorrow before merging this in, to see if anything obvious pops up.

ebin/rebar.app
fishcakez: In `rebar3` the equivalent config is `{dialyzer_plt_apps, [diameter]}`.

fishcakez:
> I didn't add tests because running them would take too long.

Could add a simple tests that just depends on the `erts` application. This should not be too slow.

ebin/rebar.app
tuncer: Unless `dialyzer_plt_apps` overrides all apps, it must be called `dialyzer_extra_apps`. Updated the patch accordingly.

tuncer:
Added new options `{dialyzer_extra_apps, []}` and `{dialyzer_plt_location, [shared, local, "custom_dir"]}`. Due to the addition of `dialyzer_plt_location`, we must always add the `_plt` suffix, as the plt is not exclusively stored in `~/.rebar/plt` anymore. Feedback welcome.

rebar.config
```diff
@@ -31,3 +31,25 @@
...
+{dialyzer_extra_apps,
+ [
+  diameter
+ ]}.
+
+{dialyzer_warnings,
+ [
+  unmatched_returns,
+  error_handling,
+  race_conditions
```
fishcakez: `race_conditions` can be slow[1]. I have not found it to be useful and avoid it because of the warning from Kostis.

[1] http://erlang.org/pipermail/erlang-bugs/2014-July/004518.html

src/rebar_dialyzer.erl
fishcakez: It might be useful to support `'_'` in other terms in the tuple because specs/types/callbacks can change between OTP releases causing subtle changes in warnings between releases.

For example the change from `queue()` to `queue:queue()` would cause the following 17.4 warning to be different in R16B03-1 and earlier.
```erlang
{warn_opaque,{"example.erl",5},{opaque_match,["pattern {A, B}","queue:queue(_)","queue:queue(_)"]}}
```

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,235 @@
...
+    ok = check_plt_existence(Plt),
+
+    Opts = [
+            {analysis_type, succ_typings},
+            {init_plt, Plt},
+            {files_rec, ["ebin"]},
+            {warnings, warnings(Config)}
+           ],
+    ?DEBUG("dialyze opts:~n~p~n", [Opts]),
+    RawWs = dialyzer:run(Opts),
```
fishcakez: There is a bug (only present in the Erlang API[1]) which can cause `dialyzer:run/1` to crash when `analysis_type` is `succ_typings` or `plt_check`. It occurs when a warning is generated due to a beam file with a warning[2] being updated since the PLT was last checked. In `rebar3` the option `{check_plt, false}` is used for `plt_check` and `succ_typings`, which avoids this issues[3], and checking the PLT is done explicitly.

[1] http://erlang.org/pipermail/erlang-bugs/2015-February/004781.html
[2] Most likely observed when a module has a callgraph warning
[3] https://github.com/rebar/rebar3/pull/206

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,235 @@
...
+'check-plt'(Config, AppFile) ->
+    {NewConfig, Plt} = plt(Config, AppFile),
+    ok = check_plt_existence(Plt),
+
+    Opts = [
+            {analysis_type, plt_check},
+            {init_plt, Plt}
+           ],
+    ?DEBUG("build-plt opts:~n~p~n", [Opts]),
+    case dialyzer:run(Opts) of
```
fishcakez: See https://github.com/rebar/rebar/pull/460/files#r26271105

Note that without the suggested fix this call will actually check the PLT twice.

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,235 @@
...
+'build-plt'(Config, AppFile) ->
+    {Config1, AppDirs} = app_dirs(Config, AppFile),
+    {NewConfig, Plt} = plt(Config1, AppFile),
+    Opts = [
+            {analysis_type, plt_build},
+            {output_plt, Plt},
+            {files_rec, AppDirs}
+           ],
+    ?DEBUG("build-plt opts:~n~p~n", [Opts]),
+    case dialyzer:run(Opts) of
```
fishcakez: This call will overwrite an existing PLT, rebuilding it from scratch. Perhaps it should `?ABORT/2` if the PLT exists and suggest deleting the PLT.

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,235 @@
...
+    {ok, NewConfig}.
+
+%% ===================================================================
+%% Internal functions
+%% ===================================================================
+
+info(help, dialyze) ->
+    info_help("Analyze the code for discrepancies");
+info(help, 'build-plt') ->
+    info_help("Build project-specific PLT");
```
fishcakez: I don't think "project-specific" is quite the correct phrase as the PLT might be shared. Perhaps just "Build the PLT" is better.

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,235 @@
...
+info(help, 'check-plt') ->
+    info_help("Check the PLT for consistency and rebuild it if it"
+              " is not up-to-date");
```
fishcakez: `check-plt` will only check the consistency of beam files already listed in the PLT. It will not add new dependencies or new beam files in existing dependencies. Also an error will be thrown if a beam file that is listed in the PLT has been deleted.

It is possible to remove "old" files using a combination of `dialyzer:plt_info/1` and `dialyzer:run/1` with analysis type `plt_remove`. New beam files can be added by using the same arguments as `build-plt` but using the analysis type `plt_add` and replacing `output_plt` with `init_plt`.

In `rebar3`[1] all beam files are found and compared with the files in the PLT using set subtract/intersections so that the minimal `remove_plt`, `plt_check` and `plt_add` analysis is done.

[1] https://github.com/rebar/rebar3/blob/ea82d0561d6366851ebcfc67488fcb7aca44179e/src/rebar_prv_dialyzer.erl#L244-L253

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,235 @@
...
+info(help, 'delete-plt') ->
+    info_help("Delete project-specific PLT").
```
fishcakez: See https://github.com/rebar/rebar/pull/460/files#r26271221.

fishcakez:
`dialyzer:run/1` can throw `{dialyzer_error, StringError}`. It might be slightly nicer to catch these throws and abort. Something like:

```erlang
try
  dialyzer:run(Opts)
catch
  throw:{dialyzer_error, Reason} -> ?ABORT("Dialyzer error: ~s", [Reason])
end
```

src/rebar_dialyzer.erl
tuncer: Are you saying the proposed use of `'_'` won't work or be of limited use?

rebar.config
tuncer: It was added to rebar's set of warnings in 2011 as suggested by @kostis, but if it's a bad idea we can change it later, as it's rebar's own dialyzer settings and therefore doesn't affect the plugin's users.

src/rebar_dialyzer.erl
tuncer: Thanks, fixed with TODO comments for a future update once Dialyzer is corrected.

src/rebar_dialyzer.erl
tuncer: Thanks, fixed.

src/rebar_dialyzer.erl
tuncer: It's supposed to, see explanation below in the remove_plt discussion.

src/rebar_dialyzer.erl
tuncer: I've updated the text, but despite possibly being shared across two copies of the same project, isn't the PLT still project-specific because it's built from and for the project?

src/rebar_dialyzer.erl
tuncer: I've considered this and as there are uncertain scenarios where we cannot calculate the differences precisely, I've decided to go the simple but always correct route:
1. `build-plt` overwrites an existing file like dialyzer_cl
2. let Dialyzer fail if the PLT doesn't match the beam files anymore

tuncer:
@fishcakez, thanks for the review, comments are inline.

Do you mind submitting a patch to properly document the `check_plt` boolean, so that we can be sure it's a public API?

src/rebar_dialyzer.erl
tuncer: I have devised a strategy for achieving this, but I'll wait for this discussion to conclude first. Maybe it's a bad idea after all.

fishcakez: Sorry I did not mean that. I meant that `'_'` could be useful in other terms, not just the line numbers. Here is the same warning as above for R16B03-1:
```erlang
{warn_opaque,{"example.erl",5},{opaque_match,["pattern {A, B}",["queue",40,41],["queue",40,41]]}}
```
The line is the same but the strings are different. To match both with one pattern, something like this is required:
```erlang
{warn_opaque,{"example.erl",5},{opaque_match,["pattern {A, B}",'_','_']}}
```

src/rebar_dialyzer.erl
fishcakez: Ah you are correct and I actually prefer your original phrasing. It is important to realise that a shared PLT will actually be specific to the project that created it because any references to fetched deps will have a path unique to the project.

src/rebar_dialyzer.erl
fishcakez: Could you explain these scenarios as it may be helpful for `rebar3`?

fishcakez:
@tuncer, sure I will do so today.

src/rebar_dialyzer.erl
tuncer: I see. It's best to leave it as a TODO and postpone the feature then until we've agreed on a design.

tuncer:
@ferd, fixup commits will be squashed once the changes pass the review.

src/rebar_dialyzer.erl
tuncer: Or we could additionally have a separate regex-ignore option that matches the formatted string.

src/rebar_dialyzer.erl
tuncer: Correct, and the idea is that the included apps change seldom enough that sharing or at least preserving the generated PLT will save time. This does of course assume the project's apps do not change frequently. If that's not the case, a local plt in `<base_dir>/.rebar/` is more useful, and you'll have to remember to run `build-plt` whenever apps change.

src/rebar_dialyzer.erl
tuncer: I'd have to reconstruct the test case, but the results were that it isn't always clear what was added/removed/updated, because the available data isn't enough to get the difference. Therefore, I've discarded the idea of a custom update-plt function. Usually `check-plt`'s update of the PLT should suffice, and if it fails a clean `build-plt` is the right choice.

src/rebar_dialyzer.erl
tuncer: Restored "project-specific".

fishcakez:
`check_plt` PR: https://github.com/erlang/otp/pull/644

tuncer:
> check_plt PR: erlang/otp#644

Thanks.

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,252 @@
...
+    {NewConfig, Plt} = plt(Config, AppFile),
+    ok = check_plt_existence(Plt),
+
+    Opts = [
+            {analysis_type, succ_typings},
+            %% http://erlang.org/pipermail/erlang-bugs/2015-February/004781.html
+            %% TODO: remove once the minimum required Erlang/OTP release includes
+            %% a Dialyzer version without the bug, or alternatively add a config
+            %% option to always check the PLT, as this may be needed by some user.
+            {check_plt, false},
```
fishcakez: Now that `{check_plt, false}` is used the PLT will not be check for consistency when running `dialyzer`, should it be checked? With `{check_plt, true}` any warnings generated when checking the PLT are ignored. To copy this behaviour there would be `plt_check` run before `succ_typings`, but ignoring any warnings from `plt_check`.

fishcakez:
It would be really nice to have the same configuration in rebar3. It is possible for rebar2 config to be forwards compatible with rebar3 config, if rebar2 can make two accommodations:

Firstly to rename `dialyzer_extra_apps` to `dialyzer_plt_extra_apps`. Given that the PLT location is configured with `dialyzer_plt_location` I  think this is reasonable. rebar3 uses 2 PLTs for faster PLT building so `dialyzer_extra_apps` would be ambiguous. I would then change some of rebar3's configuration to match the remainder of rebar2's in a forwards compatible manor. This requires adopting this PR's config and renaming `dialyzer_base_plt*` config to `dialyzer_shared_plt*` in rebar3.

Secondly there is an issue with the `dialyzer_ignored_warnings`. In rebar3 files are compiled using the absolute path, which means that file names in a dialyzer warning tuple are absolute paths and so not portable (https://github.com/rebar/rebar3/pull/272#discussion_r26578141). This could occur with custom compilers/plugins/scripts in rebar2, which means that we would need a way to recognise relative paths when the source path in a warning is absolute.

One solution to this is to replace the source file (e.g. "src/module.erl") with its base name (e.g. "module.erl") in the raw warning term. Given that this PR formats warnings using the base names this is ok in terms of Erlang modules. However this could have repercussions for non-Erlang modules where module names and source files do not align. For example Elixir does not guarantee a one-to-one mapping of modules to base names. I am unsure how much of a concern this is.

Another option is to use wildcards or regexes for file names.

Given this issue and the differences in warnings between OTP releases (https://github.com/rebar/rebar/pull/460#discussion_r26295510) I think more thought needs to go into configuring ignored warnings. Is it possible to merge this PR without this feature? I don't want to spend time changing an existing rebar3 command's config because of a new command in rebar2 only to discover the configuration won't be forwards compatible.

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,252 @@
...
+       "  ~p~n"
+       "  ~p~n"
+       "  ~p~n"
+       "  ~p~n"
+       "  ~p~n",
+       [
+        Description,
+        {dialyzer_location, shared},
+        {dialyzer_location, local},
+        {dialyzer_location, "custom_path"},
```
fishcakez: These should be `dialyzer_plt_location`.

tuncer: Thanks, fixed.

tuncer:
I'm not sure about renaming `dialyzer_extra_apps` to `dialyzer_plt_extra_apps`, as there's no other occurrence of `apps` that could make this ambiguous.

tuncer:
It's best to declare the interface for ignoring warnings preliminary and subject to change and get the rest of the plugin released. To that end, I've added a message to be printed whenever `dialyzer_ignored_warnings` is used.

tuncer:
@lrascao has suggested to put all dialyzer opts in a single `{dialyzer, ...}` tuple. We use both styles in rebar.config, and it's mostly flat. Any thoughts on that, before we merge this?

tuncer:
rebar prints absolute paths in warning and error reports only when it is necessary and tries to print relative paths where possible. This allows jumping to the error from your editor, while printing relative paths in the common case.

fishcakez:
> It's best to declare the interface for ignoring warnings preliminary and subject to change and get the rest of the plugin released. To that end, I've added a message to be printed whenever `dialyzer_ignored_warnings` is used.

rebar3 can't use this format of `dialyzer_ignored_warnings` because rebar3 has a different directory structure to rebar. For example rebar will have a file name of "src/module.erl" and rebar3 will have  "/absolute/path/_build/profiles/lib/app/src/module.erl". Therefore the file name "src/module.erl" won't match in rebar3. In order to use the same format in rebar and rebar3 we would need to use a file name format independent of directory structure. The only format completely independent of directory structure is the basename, e.g. "module.erl". Is it possible to use the base name in the raw warning? Currently this PR includes the full path in the raw warning but only the base name in the formatted warning.

> rebar prints absolute paths in warning and error reports only when it is necessary and tries to print relative paths where possible. This allows jumping to the error from your editor, while printing relative paths in the common case.

This PR uses `dialyzer:format_warning/1`, which prints the base name only. `dialyzer:format_warning/2` is required to print the "fullpath". The fullpath means the file name in the file attribute in the abstract code of the module. The fullpath might just be the basename, a relative path or an absolute path. If the path is relative it may not be relative to the current directory, only relative to the directory when the file was compiled.

fishcakez:
> I'm not sure about renaming dialyzer_extra_apps to dialyzer_plt_extra_apps, as there's no other occurrence of apps that could make this ambiguous.

Sorry I wasn't clear. This option would be ambiguous in rebar3, not rebar, and so rebar3 can't use this configuration. I was asking if it could be renamed so rebar3 can use this configuration.

fishcakez:
> @lrascao has suggested to put all dialyzer opts in a single {dialyzer, ...} tuple. We use both styles in rebar.config, and it's mostly flat. Any thoughts on that, before we merge this?

@ferd would it be possible to merge dialyzer configuration like this in rebar3? I think it would require special casing.

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,259 @@
...
+
+app_dirs(Config, AppFile) ->
+    {NewConfig, AppFileApps} = app_file_apps(Config, AppFile),
+    ?DEBUG("app file apps:~n~p~n", [AppFileApps]),
+    Deps = deps_apps(Config),
+    ?DEBUG("deps apps:~n~p~n", [Deps]),
+    ExtraApps = rebar_config:get_local(Config, dialyzer_extra_apps, []),
+    ?DEBUG("extra apps:~n~p~n", [ExtraApps]),
+    Apps = ordsets:from_list(AppFileApps ++ Deps ++ ExtraApps),
+    AppDirs = [code:lib_dir(App, ebin) || App <- ordsets:to_list(Apps)],
```
fishcakez: If the lib dir is not found this will be an error tuple, `{error, bad_name}`, which results in an unhelpful error when passed to dialyzer as a directory.

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,259 @@
...
+      Ws).
+
+app_dirs(Config, AppFile) ->
+    {NewConfig, AppFileApps} = app_file_apps(Config, AppFile),
+    ?DEBUG("app file apps:~n~p~n", [AppFileApps]),
+    Deps = deps_apps(Config),
+    ?DEBUG("deps apps:~n~p~n", [Deps]),
+    ExtraApps = rebar_config:get_local(Config, dialyzer_extra_apps, []),
+    ?DEBUG("extra apps:~n~p~n", [ExtraApps]),
+    Apps = ordsets:from_list(AppFileApps ++ Deps ++ ExtraApps),
```
fishcakez: Need to add the `erts` application into the PLT, as it is not listed in `applications` but it is an assumed dependency, otherwise functions and types in `erlang` (and other `erts` modules) will be unknown. I am unsure far back `code:lib_dir(erts, ebin)` will work, I know at least as far  back as R15B.

ferd:
@fishcakez the rule for merging in rebar3 is usually that lists get appended to each other and values get overwritten, so I believe deeply nested lists would just end up all up in each other. The process is not one I believe to be recursive. (Cc @tsloughter ?)

src/rebar_dialyzer.erl
tuncer: I think I've seen it list more than a few .app files, but if it's required to be supplied for a PLT, we add it to the set unconditionally. I'll fix it in the next update.

src/rebar_dialyzer.erl
tuncer: It's best to fail if a listed app does not exist, but we can of course terminate more gracefully. Will be fixed in the next update.

tuncer:
@fishcakez wrote:

> This PR uses dialyzer:format_warning/1, which prints the base name only. dialyzer:format_warning/2 is required to print the "fullpath". The fullpath means the file name in the file attribute in the abstract code of the module. The fullpath might just be the basename, a relative path or an absolute path. If the path is relative it may not be relative to the current directory, only relative to the directory when the file was compiled.

What's the suggested change?

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,259 @@
...
+            ok;
+        Unused ->
+            ?WARN("Unused dialyzer_ignored_warnings entries:~n~p~n", [Unused])
+    end,
+    ordsets:subtract(Ws, Patterns).
+
+print_warnings(Ws) ->
+    lists:foreach(
+      fun(W) ->
+              ?CONSOLE("~p~n~s~n", [W, dialyzer:format_warning(W)])
```
fishcakez: For full path:
```erlang
dialyzer:format_warning(W, fullpath)
```

tuncer: Okay, see 9a6a617.

src/rebar_dialyzer.erl
```diff
@@ -0,0 +1,282 @@
...
+      end,
+      Ws).
+
+format_warning(Config, W) ->
+    case rebar_utils:processing_base_dir(Config, rebar_utils:get_cwd()) of
+        true ->
+            dialyzer:format_warning(W);
+        false ->
+            dialyzer:format_warning(W, fullpath)
+    end.
```
fishcakez: I am unsure about the logic here. When processing the base directory modules will be formatted as:
```
module.erl:1: Warning
```
When `fullpath` would give:
```
src/module.erl
```
I believe the `fullpath` format is desired for the `dialyze` command in the base directory because the "full path" will be the relative path from the base director. In other cases I am unsure how/if its possible to obtain an accurate path.

tuncer: See 6d5b20f.

fishcakez:
There is still an issue with filenames in warnings:
```
==> main_app (check-plt)
src/dep_app_module.erl:3: Call to missing or unexported function erlang:bad/0
```
The file "src/dep_app_module.erl" does not exist in main_app but is at "deps/dep_app/src/dep_app_module.erl". The raw warning displayed also has the "wrong" path:
```
{warn_callgraph,{"src/dep_app_module.erl",3},{call_to_missing,[erlang,bad,0]}}
```
Dialyzer uses the `file` attribute in the abstract code of the beam file to get the source file name.

If an module from OTP generates a warning the path is always "module.erl" with no directory information. I don't think its possible to reconstruct the correct path (relative or absolute) so perhaps `filename:basename/1` should be called on the file name in the raw warnings generated by `build-plt` and `check-plt`. The paths are correct for the `dialyze` command if it is processing the base directory.

tuncer:
I haven't tried the `-dialyzer()` attribute in 18.0-rc1 yet, but maybe it's best to skip introducing filter functionality in favor of that? It sounds like the better solution to me. @ferd @asabil @fishcakez?

tuncer:
I haven't tried the `-dialyzer()` attribute in 18.0-rc1 yet, but maybe it's best to skip introducing filter functionality in favor of that. It definitely sounds like the better solution to me. @ferd @asabil @fishcakez?

fishcakez:
> I haven't tried the -dialyzer() attribute in 18.0-rc1 yet, but maybe it's best to skip introducing filter functionality in favor of that. It definitely sounds like the better solution to me. @ferd @asabil @fishcakez?

I have not tried 18.0-rc1 yet. The approach used there to filter warnings certainly looks far superior to anything that can be achieved by an external tool.

ferd:
Yeah, given this will be covered by dialyzer itself, I think we should avoid duplicating efforts and just let the new 18.x+ feature do its thing. Same for rebar3.

yannayl:
What about people who are using older versions of the OTP? (which, I assume, are the majority of users). tuncer/rebar#2 is a very simple and robust solution (4 lines of code).

ferd:
I'd say they'd be left on their own there, sadly. They would just have no supported way of filtering warnings, which I figured would not be *that* bad given right now there's no way to run Dialyzer from rebar3 to begin with.

Any feature we commit to, we kind of have to keep for good, and I'm a bit annoyed about having people  develop code and maintain it when it is clear as day that a better mechanism is going to be available and maintained by the tool itself.

Diverging opinions are welcome.

tuncer:
I agree with @ferd, because dialyzing your code can be done with 18.0, while you can of course run the code on an older VM. In fact, a newer Dialyzer usually is more capable, so it's a good idea to use the latest Dialyzer version for development.

The only issue is that of differences in the referenced code making it important to use Dialyzer against beams from an older OTP release. We have something like this with calls to functions that got removed in an OTP release where we call the old functions if the new one isn't available.

Also, the decision to ignore certain Dialyzer warnings should not be made lightly, and it seems like a good idea to isolate the relevant code into a function that's as short as possible. That way you'll have a lower chance of ignoring too much.

With all that being said, usually you shouldn't have to ignore Dialyzer warnings at all, so let's see if the attributes prove insufficient enough and that there's no way to fix it in a future Dialyzer version, before we embark on a custom filter in the plugin.

@yannayl @ferd?

tuncer:
I agree with @ferd, because dialyzing your code can be done with 18.0, while you can of course run the code on an older VM. In fact, a newer Dialyzer usually is more capable, so it's a good idea to use the latest Dialyzer version for development.

The only issue is that of differences in the referenced code making it important to use Dialyzer against beams from an older OTP release. We have something like this with calls to functions that got removed in an OTP release where we call the old functions if the new one isn't available.

Also, the decision to ignore certain Dialyzer warnings should not be made lightly, and it seems like a good idea to isolate the relevant code into a function that's as short as possible. That way you'll have a lower chance of ignoring too much.

With all that being said, usually you shouldn't have to ignore Dialyzer warnings at all, so let's see if the attribute proves insufficient enough and if there's no way to fix it in a future Dialyzer version, before we implement on a custom filter in the plugin.

@yannayl @ferd?

ferd:
Yeah, I think I'd be fine moving forward with that, similarly for rebar3.

yannayl:
Sure. If using the new Dialyzer with old VM works.

It doesn't solve my case, though, since the warnings are caused by code which is generated by the ASN1 compiler. But this is my problem.

tuncer:
> It doesn't solve my case, though, since the warnings are caused by code which is generated by the ASN1 compiler. But this is my problem.

Please report it on erlang-bugs and erlang-questions, in order for it to be corrected in the compiler. In fact, leex and yecc have been modified in 18.0-rc1 to add `-dialyzer()` attributes in the generated code, so that certain Dialyzer warnings are ignored. It should be easy to fix in the code generator.

yannayl:
Cool. Thanks.

tuncer:
See https://github.com/erlang/otp/commit/d18d8f4ba327a1c99d7812efb450b445558b154c and https://github.com/erlang/otp/commit/82a3d164aa74b8df9aec89586c9805e47c82e348.

tuncer:
If you know of warnings in code generated by snmpc, please report that as well. It would be great if the code generators in 18.0 would produce Dialyzer warning free modules.

tuncer:
If you know of warnings in code generated by snmpc, please report that as well. It would be great if all code generators in 18.0 would produce Dialyzer warning free modules. And maybe the `.idl` compiler (`ic:compile`) too.

tuncer:
So, should we nest options or not?

fishcakez:
It is possible to use nested options in rebar3. With the following config:
```
{dialyzer, [{warnings, [unmatched_returns]}]}.
{profiles, [{strict, [{dialyzer, [{warnings, [underspecs]}]}]}]}.
```
With the `strict` profile the warnings would just be `[underspecs]` to follow the rebar3 merging and not create a special case. However if it is not nested:
```
{dialyzer_warnings, [unmatched_returns]}.
{profiles, [{strict, [{dialyzer_warnings, [underspecs]}]}]}.
```
The warnings would be `[underspecs, unmatched_returns]`.

A similar case occurs for `dialyzer_plt_extra_apps` as adding an extra app to the PLT in a profile would require listing all the extra apps again.

I think this comes down to the philosophy of rebar3 config merging so I must defer to @ferd.

tuncer:
Does this mean rebar3 support flat and nested specification of dialyzer options?

tuncer:
If both are supported, it'll boil down to what's promoted and advertized as the way to do it, unless there is none.

fishcakez:
@tuncer no. It would only support one or the other. I am just showing the results of profile config merging to ask which fits with rebar3 best.

ferd:
@fishcakez A question I have is whether dialyzer tends to have you be able to entirely unset flags (override with `{flags, false}` ?)

If there is no way to turn a flag off (`{dialyzer_warnings, [{unmatched_returns, false}]}`) then it might be nicer to have nested profiles: a merged list will be able to override prior settings by just crushing previous configs, whereas a merged set of config would just keep piling on options. That does depend on whether it's a common use case. If it's not, merging lists may just be more convenient.

Of course a follow-up question is whether we'd be okay just handling this on behalf of the app, but I'd prefer not, I guess.

fishcakez:
If `unmatched_returns` is in the list it is not possible to turn it off, there is no `no_unmatched_returns`. Similar for options (e.g. `no_opaque`) that actually turn warnings off, it is not possible to turn that warning back on.

I am not keen on introducing new options in that way either. I would prefer everything was as close as possible to dialyzer itself.

Given that nesting is also more concise I am keen to just go with nesting.

tuncer:
The settings are very project (app) specific, and I doubt sharing+merging `_extra_apps` or `_warnings` will be used much.

tuncer:
Assuming we don't require nested config, this is ready to merge.

fishcakez:
Profiles are used to have multiple configurations for the same project. In the above example, dialyzer could be run with default or strict configuration for the same project. Given that an "extra_app" or "warning" can't be undone by a profile it would be better for rebar3 if config was nested. For example the strict profile can not undo a extra warning set in the default profile with the flat configuration.

tuncer:
Based only on the examples above, profiles and merging of values seem hard to use correctly and appear unintuitive to me. It reminds me of rebar's recursive command application and config inheritance issues.

I don't strongly object nesting dialyzer options, but the arguments so far seem to be about working around insufficiencies of rebar3 profiles. So, I'm leaning towards flat options.

tuncer:
Based only on the examples above, profiles and merging of values seem hard to use correctly and appear counter-intuitive to me. It reminds me of rebar's recursive command application and config inheritance issues.

I don't strongly object nesting dialyzer options, but the arguments so far seem to be about working around insufficiencies of rebar3 profiles, which most likely warrant a different fix. So, I'm leaning towards flat options.

fishcakez:
Haha, I was putting the blame purely at dialyzer itself for using a list of elements rather than proplists ;).

@tuncer: http://www.rebar3.org/v3.0/docs/profiles

tuncer:
That's also a proplist.

Another advantage of flat options is that there's no occasion for different ways to refer to an option because it's nested. Granted it's a minor issue, but I think that's easier for users to discover. If you want to pass different options to a single function, a proplist makes a lot of sense, and in rebar's case the list of terms returned from a config file is that proplist for rebar:run.

fishcakez:
The warnings are not a proplist and that is the problem. `{unmatched_returns, false}` does not remove `unmatched_returns`.

fishcakez:
Oh wait I misunderstood. The options passed to `dialyzer:run/1` are a proplist and that is why it is better for rebar3 to used nested config: `{dialyzer, Proplist}` because then the proplist works like `dialyzer:run/1`.

tuncer:
> The warnings are not a proplist and that is the problem. `{unmatched_returns, false}` does not remove `unmatched_returns`.

You wouldn't want rebar to modify values at that level anyway.

tuncer:
The warnings are the only bit that's passed as is to `dialyzer:run/1`, while the other options are not.

I have re-reviewed the example, and merging as described still looks counter-intuitive.

fishcakez:
Rebar3 has `get_warnings` and `plt` options too.

I picked the example to show handling of a difficult case so it may not be intuitive.

In `rebar3` with nested config the options would be merged as so:
```erlang
{dialyzer, Proplist2 ++ Proplist1}.
```
With flat:
```erlang
{dialyzer_warnings, WarningFlags2 ++ WarningFlags1}.
```
I think both of these are intuitive as the merge is just appending a list to a list.

However the problem with flat configuration is that the warning flags in one profile are forced onto another profile. This means that having a warning flag in the default profile forces all profiles to have that warning flag.

The issue doesn't occur for nested config because the profile overrides the default.

If it is nested should it be `dialyzer_opts` to match other config?

ferd:
The merge is more complex than that. It's using a proplist sort and them merging those. but the practical end result would be logically similar to what you mention.

fishcakez:
Ah thank you, @ferd.

tuncer:
Merging is highly value-dependent, and there is not one true way to merge. Therefore, the user needs control over the merge process. If the user had control, there would be no issue with dialyzer options.

tuncer:
Merging is highly value-dependent, and there is no one true way to merge. Therefore, the user needs control over the merge process. If the user had control, there would be no issue with dialyzer options.

fishcakez:
Just did a fresh pull and plugin fails to build a PLT if `~/.rebar/plt/` does not exist. It's possible this always occurred but the directory existed so I didn't notice.
```
==> project (build-plt)
ERROR: Dialyzer error:
    The PLT file /home/fish/.rebar/plt/project_17.4_plt is not writable
ERROR: 'build-plt' failed while processing /home/fish/project: rebar_abort
```
Can `filelib:ensure_dir/1` be used to create the directory if it does not exist?

tuncer:
Will fix it tomorrow.

tuncer:
Fixed.

rebar.config
```diff
@@ -31,3 +31,25 @@
...
+ ]}.
+
+{dialyzer_ignored_warnings,
```
fishcakez: No ignored warnings anymore, otherwise :+1: assuming config is remaining flat.

tuncer: Fixed.

PR #464, opened 2015-03-18
Fix error reporting crash introduced in 2e9706f

c-bik:
Crash:
```erlang
ERROR: compile failed while processing c:/projects/git/sqlparse: {'EXIT',{badarg,[{io,format,
                     [<0.23.0>,"Compiling ~s failed:\n",
                      [{"src/somemodule.yrl","src/somemodule.erl"}]],
                     []},
                 {rebar_base_compiler,compile_queue,3,
                                      [{file,"src/rebar_base_compiler.erl"},
                                       {line,183}]},
                 {rebar_erlc_compiler,compile,2,
                                      [{file,"src/rebar_erlc_compiler.erl"},
                                       {line,99}]},
                 {rebar_core,run_modules,4,
                             [{file,"src/rebar_core.erl"},{line,491}]},
                 {rebar_core,execute,6,
                             [{file,"src/rebar_core.erl"},{line,416}]},
                 {rebar_core,maybe_execute,8,
                             [{file,"src/rebar_core.erl"},{line,300}]},
                 {rebar_core,process_dir1,7,
                             [{file,"src/rebar_core.erl"},{line,259}]},
                 {rebar_core,process_commands,2,
                             [{file,"src/rebar_core.erl"},{line,91}]}]}}
```

After fix:
```erlang
Compiling src/somemodule.yrl failed:
```

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -134,3 +134,4 @@ stwind
 Pavel Baturko
 Igor Savchuk
 Mark Anderson
+Bikram Chatterjee
--- a/src/rebar_base_compiler.erl
+++ b/src/rebar_base_compiler.erl
@@ -278,7 +278,7 @@ format_error(AbsSource, Extra, {Mod, Desc}) ->
 maybe_absname(Config, Filename) ->
     case rebar_utils:processing_base_dir(Config) of
         true ->
-            Filename;
+            unit_source(Filename);
         false ->
-            filename:absname(Filename)
+            filename:absname(unit_source(Filename))
     end.


tuncer:
Looks good to me. @lrascao?

Before we can merge this, can you please fix the commit message following [CONTRIBUTING#writing-commit-messages](https://github.com/rebar/rebar/blob/master/CONTRIBUTING.md#writing-commit-messages). Maybe `Fix error reporting crash introduced in 2e9706f`?

tuncer:
Also, you may want to append your name in `THANKS`.

tuncer:
Also, you may want to append your name in `THANKS` in the same commit.

c-bik:
@tuncer Thanks. And sorry about not following pull request guidelines.
Hopefully I got it right this time!

lrascao:
maybe @tomas-abrahamsson would also like to take a look at it.
@c-bik, on compile_each/3, maybe_absname is being invoked with the 2nd parameter using the output from unit_source, was there an issue applying the same reasoning to other maybe_absname invocations?

c-bik:
@lrascao, I see maybe_absname from compile_each, compile_queue and format_errors, only from compile_each passes on from unit_source. Wondering if its best to remove that now from compile_each call too, now that unit_source is called in maybe_absname for all scenarios.

lrascao:
yeah, i guess both approaches are acceptable, however @tomas-abrahamsson's input is the most relevant to determine if there are no side-effects lurking

tuncer:
@tomas-abrahamsson?

lrascao:
Interestingly, this PR solves the exact some issue found and fixed in #491, the aproach is different though, this fix is made at the level of the maybe_abs_name method.
#491 was one level higher in compile_queue, consistency is kept with compile_each that has the same aproach

ferd:
Then I guess this one here is good to close?

lrascao:
IMO yes, however we should have used this PR as the base for the fix instead of creating a new one, my fault for not noticing that they were the same problem...

Issue #357, opened 2014-09-06
Parse transform in rebar config doesn't work


kejv:
First, let me present a motivation for having a parse transform in rebar config file. In our company we slowly start adapting to erlang r17 and we have to solve an issue which I guess almost eveybody has to deal with: backwards incompatible change of dict(), set(), etc. types. Approach taken here in rebar project isn't an option for us since we use those types in many modules and also don't want to have that version-conditioned macro in every new module. We think that much more transparent approach is to apply parse transform which just modifies the types according to given version. This means that we want to apply parse transform to every module (even the future ones) and have this stated only in one place: rebar.config (as opposed to having to write parse_transform compile option to every single module). 

So, in order that parse transform works correctly, it has to be compiled befory any "other" modules, i.e. in a way similar to erl_first_files. I have that almost fixed in my fork but I still see couple of problems and I want them solved before I send a pull request:
- Modules from subprojects (dependencies) are compiled before modules from main project. But the subprojects also inherit main project's rebar config which means that parse transform is applied to subprojects before it is actually compiled, which is of course an error. I think that in this situation, though, it makes sense just to ignore any parse transforms for subprojects.
- Second one is tougher. Which modules from main project should be compiled first: parse transform or erl_first_files? I can imagine that both possible orders make sense so I would like to start a discussion on this topic here.

I am using most recent version of rebar as of today (#9583d45).

lrascao:
Closing as both #466 and #477 have been merged

PR #466, opened 2015-03-22
Improve test targets in Makefile

kejv:
- enable running eunit and retest tests separately
- add neccessary dependencies to each test target so that
  one can just do 'make test' without recompiling changed files
  manually
- enable overwriting retest's log level
- use retest's own Makefile instead of invoking rebar manually

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,11 @@
-.PHONY: clean dialyzer_warnings xref_warnings deps test
+.PHONY: clean dialyzer_warnings xref_warnings deps test test_eunit test_inttest
 
 REBAR=$(PWD)/rebar
 RETEST=$(PWD)/deps/retest/retest
 OTPVSNCMD='io:fwrite("~s",[rebar_utils:otp_release()]), halt().'
 OTPVSN=$(shell erl -pa ebin/ -noshell -eval $(OTPVSNCMD))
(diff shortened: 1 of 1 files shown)


ferd:
I'm fine with this. @tuncer ?

Makefile
```diff
@@ -64,10 +65,14 @@ binary: clean all
 
 deps:
 	@REBAR_EXTRA_DEPS=1 ./rebar get-deps
-	@(cd deps/retest && $(REBAR) compile escriptize)
+	$(MAKE) -C deps/retest debug
```
tuncer: `retest` doesn't have to built in debug mode.

tuncer:
Other than the unneeded debug build of retest, +1.

tuncer:
@kejv you may want to append your name in `THANKS` as part of this commit.

kejv:
I would wait with adding my name to THANKS, since I am currently working on some bigger refactor as part of #357, and then you will see if it's worth thanking me or not :-)

Makefile
kejv: I thought so. I have just put the debug there, so it's equvivalent with the previous behaviour (see rebar.config in deps/retest). I will remove it in a fixup.

tuncer:
+1, but please don't forget to squash the fixup first.

kejv:
Ok. Ready to merge.

PR #467, opened 2015-03-29
Refactor logic and optimizations in rebar_erlc_compiler:doterl_compile/4

kejv:
- Change format of erlcinfo so that it is more robust and contains enough info for deciding which files need to be recompiled.
- Add tests.
- Remove unneeded code.

### DIFF
--- a/src/rebar_erlc_compiler.erl
+++ b/src/rebar_erlc_compiler.erl
@@ -398,20 +398,17 @@ needs_compile(Source, Target, Parents) ->
     lists:any(fun(I) -> TargetLastMod < filelib:last_modified(I) end,
               [Source] ++ Parents).
 
-check_erlcinfo(#erlcinfo{vsn=?ERLCINFO_VSN}) ->
-    ok;
-check_erlcinfo(#erlcinfo{vsn=Vsn}) ->
-    ?ABORT("~s file version is incompatible. expected: ~b got: ~b~n",
-           [erlcinfo_file(), ?ERLCINFO_VSN, Vsn]);
-check_erlcinfo(_) ->
-    ?ABORT("~s file is invalid. Please delete before next run.~n",
-           [erlcinfo_file()]).
-
 erlcinfo_file() ->
     filename:join([rebar_utils:get_cwd(), ".rebar", ?ERLCINFO_FILE]).
 
 init_erlcinfo(InclDirs, Erls) ->
-    G = restore_erlcinfo(),
+    G = digraph:new(),
+    try restore_erlcinfo(G)
+    catch
+        _ ->
+            ?WARN("Failed to restore ~s file. Discarding it.~n", [erlcinfo_file()]),
+            ok = file:delete(erlcinfo_file())
+    end,
     %% Get a unique list of dirs based on the source files' locations.
     %% This is used for finding files in sub dirs of the configured
     %% src_dirs. For example, src/sub_dir/foo.erl.
@@ -460,34 +457,21 @@ modify_erlcinfo(G, Source, Dirs) ->
               digraph:add_edge(G, Source, Incl)
       end, AbsIncls).
 
-restore_erlcinfo() ->
-    G = digraph:new(),
+restore_erlcinfo(G) ->
     case file:read_file(erlcinfo_file()) of
         {ok, Data} ->
-            try binary_to_term(Data) of
-                Erlcinfo ->
-                    ok = check_erlcinfo(Erlcinfo),
-                    #erlcinfo{info=ErlcInfo} = Erlcinfo,
-                    {Vs, Es} = ErlcInfo,
-                    lists:foreach(
-                      fun({V, LastUpdated}) ->
-                              digraph:add_vertex(G, V, LastUpdated)
-                      end, Vs),
-                    lists:foreach(
-                      fun({V1, V2}) ->
-                              digraph:add_edge(G, V1, V2)
-                      end, Es)
-            catch
-                error:badarg ->
-                    ?ERROR(
-                       "Failed (binary_to_term) to restore rebar info file."
-                       " Discard file.~n", []),
-                    ok
-            end;
-        _Err ->
+            #erlcinfo{vsn=?ERLCINFO_VSN, info={Vs, Es}} = binary_to_term(Data),
+            lists:foreach(
+              fun({V, LastUpdated}) ->
+                      digraph:add_vertex(G, V, LastUpdated)
+              end, Vs),
+            lists:foreach(
+              fun({V1, V2}) ->
+                      digraph:add_edge(G, V1, V2)
+              end, Es);
+        {error, _} ->
             ok
-    end,
-    G.
+    end.
 
 store_erlcinfo(_G, _Modified = false) ->
     ok;
--- a/src/rebar_erlc_compiler.erl
+++ b/src/rebar_erlc_compiler.erl
@@ -672,14 +672,8 @@ parse_attrs(Fd, Includes) ->
     end.
 
 process_attr(Form, Includes) ->
-    try
-        AttrName = erl_syntax:atom_value(erl_syntax:attribute_name(Form)),
-        process_attr(AttrName, Form, Includes)
-    catch _:_ ->
-            %% TODO: We should probably try to be more specific here
-            %% and not suppress all errors.
-            Includes
-    end.
+    AttrName = erl_syntax:atom_value(erl_syntax:attribute_name(Form)),
+    process_attr(AttrName, Form, Includes).
 
 process_attr(import, Form, Includes) ->
     case erl_syntax_lib:analyze_import_attribute(Form) of
@@ -719,8 +713,12 @@ process_attr(compile, Form, Includes) ->
                       [atom_to_list(M) ++ ".erl"|Acc];
                  (_, Acc) ->
                       Acc
-              end, Includes, L)
-    end.
+              end, Includes, L);
+        _ ->
+            Includes
+    end;
+process_attr(_, _Form, Includes) ->
+    Includes.
 
 %% Given the filename from an include_lib attribute, if the path
 %% exists, return unmodified, or else get the absolute ERL_LIBS
--- a/src/rebar_erlc_compiler.erl
+++ b/src/rebar_erlc_compiler.erl
@@ -40,11 +40,11 @@
 -define(ERLCINFO_FILE, "erlcinfo").
 -type erlc_info_v() :: {digraph:vertex(), term()} | 'false'.
 -type erlc_info_e() :: {digraph:vertex(), digraph:vertex()}.
--type erlc_info() :: {list(erlc_info_v()), list(erlc_info_e())}.
+-type erlc_info() :: {list(erlc_info_v()), list(erlc_info_e()), list(string())}.
 -record(erlcinfo,
         {
           vsn = ?ERLCINFO_VSN :: pos_integer(),
-          info = {[], []} :: erlc_info()
+          info = {[], [], []} :: erlc_info()
         }).
 
 -ifdef(namespaced_types).
@@ -387,10 +387,6 @@ u_add_element(Elem, [Elem|_]=Set) -> Set;
 u_add_element(Elem, [E1|Set])     -> [E1|u_add_element(Elem, Set)];
 u_add_element(Elem, [])           -> [Elem].
 
--spec include_path(file:filename(), list()) -> [file:filename(), ...].
-include_path(Source, InclDirs) ->
-    lists:usort(["include", filename:dirname(Source) |InclDirs]).
-
 -spec needs_compile(file:filename(), file:filename(),
                     [string()]) -> boolean().
 needs_compile(Source, Target, Parents) ->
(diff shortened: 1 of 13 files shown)


tuncer:
Thanks for the patch.

@norton and @nevar, can check the changes don't break your projects?

Please do not rewrite the branch for any followup fixes, because it's a large change and, given GitHub's review system, it's the only way to review further changes. That means, make any fixes in extra commits, even though it's many commits.

Can you separately explain the include_path and store_erlc_info changes in more detail, in order to leave no room for interpretation? If it makes sense, maybe as a code comment.

* 971aa39 OK
* 4d9a4f7 OK
* 96f5c9f OK
* eade8e3 WHYNOT but:
 * it must be `ok = file:delete()`
 * we have so selectively catch just `error:badarg`
* 7ebbbf5 OK
* a1f7dee OK but:
 * erlc_dep_graph_rt.erl may not use os:cmd but has to use `file:read_file_info/1`, `file:change_time/2`, and `rebar_file_utils:mv/2`
 * isn't the timer:sleep(1000) bound to sometimes fail? can't we do better than that?
 * comments must be %% and not %, unless they're trailing in a code line
 * shouldn't `fun erlang :'<'/2` be `fun erlang:'<'/2`?
* 4df0c83 NEEDINFO
 * That catch was in place since the original speedup refactoring, because it was impossible to know beforehand which exceptions are safe to ignore. How has this changed, that we can let them all through now?
* 64ca0d3 NEEDINFO
 * explain include_path change in more detail
* 1243359 NEEDINFO
 * we have to match `ok = store_erlcinfo()`, so don't turn `modified|unmodified` into a boolean
* 181b0fe NEEDINFO
 * comments must be %% not %
* a2ca6d8 NEEDINFO
 * reasoning makes sense, but do you have measurements for reference?
* 4c4cce4 OK (to check by Joe Norton use case)
* c244c84 OK
 * We cannot rely on `sed`. either use a separate file or make the modification with Erlang functions.
* b86a07e NEEDINFO
 * comments must be %%, not %
 * just curious, any particular reason for naming the function in perl.erl such that it has to be quoted?
 * typo "alread respect" -> "already respect"
 * in `needed_files/3`, what does `[Source||Needed]` do?
 * in `compile_xrl_yrl/5`, why not introduce a replacement `needs_compile/2` instead of inlining the code?
 * can't we restore the debug logging of dependencies, or do you think it's irrelevant and hard to log in the new version?

tuncer:
Thanks for the patch.

@norton and @nevar, can you check the changes don't break your projects?

You forgot to append your name in `THANKS`.

Please do not rewrite the branch for any followup fixes, because it's a large change and, given GitHub's review system, it's the only way to review further changes. That means, make any fixes in extra commits, even though it's many commits.

Can you separately explain the include_path and store_erlc_info changes in more detail, in order to leave no room for interpretation? If it makes sense, maybe as a code comment.

* 971aa39 OK
* 4d9a4f7 OK
* 96f5c9f OK
* eade8e3 WHYNOT but:
 * it must be `ok = file:delete()`
 * we have so selectively catch just `error:badarg`
* 7ebbbf5 OK
* a1f7dee OK but:
 * erlc_dep_graph_rt.erl may not use os:cmd but has to use `file:read_file_info/1`, `file:change_time/2`, and `rebar_file_utils:mv/2`
 * isn't the timer:sleep(1000) bound to sometimes fail? can't we do better than that?
 * comments must be %% and not %, unless they're trailing in a code line
 * shouldn't `fun erlang :'<'/2` be `fun erlang:'<'/2`?
* 4df0c83 NEEDINFO
 * That catch was in place since the original speedup refactoring, because it was impossible to know beforehand which exceptions are safe to ignore. How has this changed, that we can let them all through now?
* 64ca0d3 NEEDINFO
 * explain include_path change in more detail
* 1243359 NEEDINFO
 * we have to match `ok = store_erlcinfo()`, so don't turn `modified|unmodified` into a boolean
* 181b0fe NEEDINFO
 * comments must be %% not %
* a2ca6d8 NEEDINFO
 * reasoning makes sense, but do you have measurements for reference?
* 4c4cce4 OK (to check by Joe Norton use case)
* c244c84 OK
 * We cannot rely on `sed`. either use a separate file or make the modification with Erlang functions.
* b86a07e NEEDINFO
 * comments must be %%, not %
 * just curious, any particular reason for naming the function in perl.erl such that it has to be quoted?
 * typo "alread respect" -> "already respect"
 * in `needed_files/3`, what does `[Source||Needed]` do?
 * in `compile_xrl_yrl/5`, why not introduce a replacement `needs_compile/2` instead of inlining the code?
 * can't we restore the debug logging of dependencies, or do you think it's irrelevant and hard to log in the new version?

kejv:
> Please do not rewrite the branch for any followup fixes, because it's a large change and, given GitHub's review system, it's the only way to review further changes. That means, make any fixes in extra commits, even though it's many commits.

Don't worry, I am used to doing review by sending just fixups. I have made a force push shortly after initial pull request only because nobody had probably looked at it at that time. Although here the fixups will bring some merge conflicts during final rebase, but I should manage.

Actually I am also used to commenting on particular lines of source code instead of writing comments to conversation sections. So if you don't mind I will continue the review in particular commits, where appropriate.

kejv:
I think a have adressed all your comments. Either by sending a fixup or by commenting to corresponding commit.

norton:
@tuncer Here is the result with one project - ubf.  I did not check the details of the failure.

```
$ erl --version
Erlang/OTP 17 [erts-6.3] [source] [64-bit] [smp:4:4] [async-threads:10] [hipe] [kernel-poll:false]

Eshell V6.3  (abort with ^G)
1> 

$ git clone git@chibiko.dyndns.org:ubf.git

$ cd ubf

$ cp <<path-to-rebar>> ./rebar

$ ./rebar --version
rebar 2.5.1 17 20150330_022056 git 2.5.1-165-g82a1756

$ ./rebar get

$ make 
./rebar compile
==> qc (compile)
ERROR: compile failed while processing /Users/norton/tmp/ubf/deps/qc: {'EXIT',
    {function_clause,
        [{filename,join,
             [{error,bad_name},"triq_statem.hrl"],
             [{file,"filename.erl"},{line,406}]},
         {rebar_erlc_compiler,process_attr,3,
             [{file,"src/rebar_erlc_compiler.erl"},{line,591}]},
         {rebar_erlc_compiler,parse_attrs,2,
             [{file,"src/rebar_erlc_compiler.erl"},{line,559}]},
         {rebar_erlc_compiler,modify_erlcinfo,4,
             [{file,"src/rebar_erlc_compiler.erl"},{line,389}]},
         {rebar_erlc_compiler,'-modify_erlcinfo/4-fun-0-',4,
             [{file,"src/rebar_erlc_compiler.erl"},{line,396}]},
         {lists,foreach,2,[{file,"lists.erl"},{line,1336}]},
         {rebar_erlc_compiler,modify_erlcinfo,4,
             [{file,"src/rebar_erlc_compiler.erl"},{line,394}]},
         {rebar_erlc_compiler,'-modify_erlcinfo/4-fun-0-',4,
             [{file,"src/rebar_erlc_compiler.erl"},{line,396}]}]}}
make: *** [compile] Error 1
```

tuncer:
> Don't worry, I am used to doing review by sending just fixups. I have made a force push shortly after initial pull request only because nobody had probably looked at it at that time. Although here the fixups will bring some merge conflicts during final rebase, but I should manage.

I don't mind having the fixups as extra commits, though you'd have to fix the commit message in that case.

> Actually I am also used to commenting on particular lines of source code instead of writing comments to conversation sections. So if you don't mind I will continue the review in particular commits, where appropriate.

Of course, it's just that Github's review system is insufficient, and I need a way to keep track of review progress. There's a lot of commits, but I'll try to not miss any of the inline comments.

src/rebar_erlc_compiler.erl
```diff
@@ -292,144 +284,74 @@ doterl_compile(Config, OutDir) ->
...
-            ulist(T, [H|Acc])
-    end;
-ulist([], Acc) ->
-    lists:reverse(Acc).
+%% Get subset of SourceFiles which need to be recompiled, respecting
+%% dependencies induced by given graph G.
+needed_files(G, OutDir, [Source|Rest]) ->
+    Target = target_base(OutDir, Source) ++ ".beam",
+    Needed = digraph:vertex(G, Source) > {Source, filelib:last_modified(Target)},
+    [Source||Needed] ++ needed_files(G, OutDir, Rest);
```
nevar: Why not:
```erlang
needed_files(G, OutDir, ErlFileList) ->
    [File ||
        File <- ErlFileList,
        digraph:vertex(G, File) > {File, beam_last_modified(OutDir, File)}].

beam_last_modified(OutDir, Source) ->
    filelib:last_modified(target_base(OutDir, Source) ++ ".beam").
```

nevar:
Great work. Looked code. Not much only part related to ``erl_first_files``.

src/rebar_erlc_compiler.erl
kejv: I have solved it very similarly with `lists:filter` as per suggestion by @ferd.

kejv:
So I think that I have adressed all issues, except that `andalso` thing in 1243359, which I will repair in fixup during rebase. So if anyone has any comments to particular commits (or fixups), I will wait cca 24 hours and then I will squash all fixups together, therefore all comments to current commits will be lost. Then I want to recheck the issue reported by @norton, although I don't have a clue how to debug it yet. If ubf is a private repository, can you give some hints what possible unusual you might be doing there?

norton:
ubf is not a private repository.

tuncer:
Let's get the regression(s) fixed and move this forward. We can address subjective code issues independently.

kejv:
I have squashed all the fixups, fixed merge conflicts along the way and added one additional fixup for that `andalso` thing as agreed (look also how it affects code in ef0165f). Hopefully everything should be ok now.

kejv:
@norton I have looked at the problem you reported and it really is caused by d61b51b. At that commit I have stopped silently supressing some errors from process_attr/3 and this one of yours suddenly popped up (it is present in current master). I will send a fix with explanation.

tuncer:
@kejv wrote:
> @norton I have looked at the problem you reported and it really is caused by d61b51b. At that commit I have stopped silently supressing some errors from process_attr/3 and this one of yours suddenly popped up (it is present in current master). I will send a fix with explanation.

Yeah, I found+fixed the initial problems with include_lib only after disabling the catch-all, and then commented it shouldn't unconditionally suppress all errors. I like that we don't do that anymore, and I wonder if we have a means of reporting any other error in a nicer way. I'm thinking of anything that might throw in `AttrName = erl_syntax:atom_value(erl_syntax:attribute_name(Form))`. We probably cannot, but it doesn't hurt to consider our options.

Other than that, this is ready to merge, isn't it?

kejv:
>  I wonder if we have a means of reporting any other error in a nicer way. I'm thinking of anything that might throw in AttrName = erl_syntax:atom_value(erl_syntax:attribute_name(Form))

As I said earlier I have experimented a little with syntax errors in compile attributes, hoping that I will get exception here in process_attr/2 instead of in actual compilation, but I didn't (at least the error messages were the same as erlc). So if there's a source of rebar exceptions here I don't of it.

tuncer:
It would be easier to match `ok | {error, Reason}`, but the way erl_syntax's api works (see [erlang-questions thread](http://erlang.org/pipermail/erlang-questions/2014-April/078788.html), we can not know what exception it may throw. There may be other code out there that will trigger an exception here, and if we can present such an error in a more user friendly way, we should do it. That said, I don't consider this a merge blocker, so it's fine to merge as is.

tuncer:
It would be easier to match `ok | {error, Reason}`, but the way erl_syntax's api works (see [erlang-questions thread](http://erlang.org/pipermail/erlang-questions/2014-April/078788.html)), we can not know what exception it may throw. There may be other code out there that will trigger an exception here, and if we can present such an error in a user friendlier way, we should do it. That said, I don't consider this a merge blocker, so it's fine to merge as is.

kejv:
Ok then. I will do one final rebase and then it can be merged.

norton:
@kejv changes look good :+1:  Thank you.

```
$ ./rebar --version
rebar 2.5.1 17 20150406_222000 git 2.5.1-174-ge17d30a
```

tuncer:
@ferd, okay to merge?

ferd:
Let's see how it goes and if it breaks stuff!

carlosedp:
I saw the same problem reported on #479 where my project uses "include_lib" statements when I updated rebar.
This one works perfecly:
`> ./rebar --version
rebar 2.5.1 17 20150324_112243 git 2.5.1-141-g02de46d-dirty`

This one gives me error:
`> ./rebar --version
rebar 2.5.1 17 20150422_145452 git 2.5.1-181-g56ac194-dirty`

The error is:
`==> dccaserver (compile)
ERROR: compile failed while processing e:/cygwin17/home/cpaula/dcca-server-OTP/apps/dccaserver: {'EXIT',
    {function_clause,
        [{filename,join,[[]],[{file,"filename.erl"},{line,392}]},
         {rebar_erlc_compiler,expand_include_lib_path,1,
             [{file,"src/rebar_erlc_compiler.erl"},{line,647}]},
         {rebar_erlc_compiler,process_attr,3,
             [{file,"src/rebar_erlc_compiler.erl"},{line,597}]},
         {rebar_erlc_compiler,parse_attrs,2,
             [{file,"src/rebar_erlc_compiler.erl"},{line,565}]},
         {rebar_erlc_compiler,modify_erlcinfo,4,
             [{file,"src/rebar_erlc_compiler.erl"},{line,400}]},
         {rebar_erlc_compiler,'-update_erlcinfo_fun/2-fun-0-',4,
             [{file,"src/rebar_erlc_compiler.erl"},{line,359}]},
         {lists,foldl,3,[{file,"lists.erl"},{line,1261}]},
         {rebar_erlc_compiler,init_erlcinfo,2,
             [{file,"src/rebar_erlc_compiler.erl"},{line,349}]}]}}
Makefile:25: recipe for target 'compile' failed
make: *** [compile] Error 1`

tuncer:
@carlosedp see #485

PR #473, opened 2015-04-05
Add options 'groups' and 'cases' to 'ct' command

RubberCthulhu:
* Add an option `groups` to `rebar ct` command. So the command 
`rebar ct suites=Suite1 groups=Group1,Group2,...,GroupN`
is equal to 
`ct_run -suite Suite1_SUITE -group Group1 Group2 ... GroupN` 
It allows to run specified test groups in specified test suite with Common Test tool. Besides it is absolutely necessary to specify groups for running test cases which are included in these groups, otherwise `init_per_group/2` and `end_per_group/1` callbacks are not called by `ct_run`.

* Add an option `cases` to `rebar ct` command. So the command 
`rebar ct suites=Suite1 cases=Case1,Case2,...,CaseN` 
is equal to 
`ct_run -suite Suite1_SUITE -case Case1 Case2 ... CaseN` 
It allows to run one or more test cases in specified test suite. Currently `rebar` has an option `case` which allows to run only one test case. The option `case` is remained for backward compability. It's suggested to consider the option `case` as deprecated and is recommended to use `cases` instead.

* Update help messages according to new command line options.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -135,3 +135,4 @@ Pavel Baturko
 Igor Savchuk
 Mark Anderson
 Brian H. Ward
+Danil Onishchenko
--- a/src/rebar.erl
+++ b/src/rebar.erl
@@ -450,7 +450,7 @@ eunit       [suite[s]=foo]               Run EUnit tests in foo.erl and
             [random_suite_order=Seed]    with a random seed for the PRNG, or a
                                          specific one.
 
-ct          [suite[s]=] [case=]          Run common_test suites
+ct          [suite[s]= [group[s]= [case[s]=]]] Run common_test suites
 
 qc                                       Test QuickCheck properties
 
--- a/src/rebar_ct.erl
+++ b/src/rebar_ct.erl
@@ -67,8 +67,16 @@ info(help, ct) ->
        "  ~p~n"
        "  ~p~n"
        "Valid command line options:~n"
-       "  suites=foo,bar - run <test>/foo_SUITE and <test>/bar_SUITE~n"
-       "  case=\"mycase\" - run individual test case foo_SUITE:mycase~n",
+       "  suites=Suite1,Suite2,...,SuiteN~n"
+       "      - run Suite1_SUITE, Suite2_SUITE, ..., SuiteN_SUITE~n"
+       "      in the test folder.~n"
+       "  groups=Group1,Group2,...,GroupN~n"
+       "      - run test groups Group1, Group2, ..., GroupN of specified suites.~n"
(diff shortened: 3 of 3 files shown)


ferd:
Sounds reasonable to me.

tuncer:
@ferd wrote:
> Sounds reasonable to me.

Yes.

ferd:
Ah yeah. I don't think we should really deprecate much in that way. I had forgotten about `case=`.

tuncer:
@ferd to be clear, are you saying we should deprecate?

ferd:
I'm saying we can't really do that, so that's a bit of a blocker.

RubberCthulhu:
@ferd
Sorry, but it's not quite clear what you mean.

Perhaps my explanation was a bit confusing. My point is:
The option `case` remains available for `rebar ct`, hence it can still be used. However I suggest to consider `case` as out-of-date and recommend to use `cases` instead because it's more generic.

ferd:
As long as the old option keep working, then yeah, we can make that work.

What happens if both options are used at once, does the new one just take precedence or do you make the union of both sets?

RubberCthulhu:
The new one take precedence. It seems reasonable to me - if you begin to use new option then why do you need the old one?

ferd:
Yeah. It makes sense to me too. This question is just a good way to figure out if many corner cases have been considered (like in using case= and -cases at the same time).

RubberCthulhu:
@ferd @tuncer 
So, should I add a message which informs that `cases` is preferable to use instead `case`?

Are there any other remarks about the patch?
Sorry to have bothered you, but it's not clear for me if the patch is considered as ready for merge or it still requires some additions.

tuncer:
@ferd?

ferd:
A message that says if it's preferable to use the new one works, as long as the old one doesn't stop working.

Sorry I get entirely swamped across projects and work and there's stuff I let slip through the cracks :(

tuncer:
You can use rebar_utils:get_deprecated_* for that.

RubberCthulhu:
@tuncer @ferd 
I've added the message for old options via rebar_utils:get_deprecated_global.

RubberCthulhu:
@tuncer 
The defects you listed have been fixed.

tuncer:
On Mon, May 25, 2015 at 4:19 PM, Danil Onishchenko
<notifications@github.com> wrote:
>
> @tuncer
> The defects you listed have been fixed.

+1, and feel free to collapse the fixup commits

RubberCthulhu:
Done.

PR #475, opened 2015-04-08
Adapt to 18.x time api changes

tuncer:


tuncer:
This does the right thin on 18.x, but I don't know how to silence the compiler warning, and I'm not sure why it's reported at all, since bootstrap is not supposed to be compiled before execution.

tuncer:
This does the right thing on 18.x, but I don't know how to silence the compiler warning, and I'm not sure why it's reported at all, since bootstrap is not supposed to be compiled before execution.

ferd:
There's actually a compatibility module that the OTP team published that might be interesting to adopt, although it does rely on `try...catch` to work and would be subject to similar warnings.

Otherwise we might have to go the macro way like for typespecs.

tuncer:
> There's actually a compatibility module that the OTP team published that might be interesting to adopt, although it does rely on try...catch to work

I prefer this style, because it's self-documenting, whereas a try-catch may not be as obvious, but what are the details of that module?

> and would be subject to similar warnings.

Is it a header where you could define a macro to silence the warning?

In any case, I don't know why escript complains here, given we didn't tell it to compile bootstrap. We should figure that out, and maybe it's a bug.

> Otherwise we might have to go the macro way like for typespecs.

Are you suggesting to introduce an indirection to execute the bootstrap escript?

tuncer:
Added `-compile({nowarn_deprecated_function, [{erlang, now, 0}]})`, but that shouldn't be necessary, since we don't have `-mode(compile)`.

bootstrap
```diff
@@ -121,10 +121,21 @@ rm(Path) ->
...
 
 build_time() ->
-    {{Y, M, D}, {H, Min, S}} = calendar:now_to_universal_time(now()),
+    {{Y, M, D}, {H, Min, S}} = calendar:now_to_universal_time(rebar_now()),
     lists:flatten(io_lib:format("~4..0w~2..0w~2..0w_~2..0w~2..0w~2..0w",
                                 [Y, M, D, H, Min, S])).
 
+%% Even though we don't have -mode(compile), erl_lint:module/1 is run in
+%% escript, so we still have to silence the deprecation warning.
+-compile({nowarn_deprecated_function, [{erlang, now, 0}]}).
```
fishcakez: When using `nowarn_deprecated_function` the functions should only be included if the functions are deprecated for all supported versions. Otherwise a warning is raised in versions in releases where the function is not deprecated.

tuncer: That's right, but we'd have to add a wrapper around bootstrap, or do you know of another solution?

tuncer: That's right, but we'd have to add a wrapper around bootstrap or run bootstrap from bootstrap. Unless you know a simpler solution, I'm not sure it's worth the complexity.

fishcakez: It can be used without naming the function:
```erlang
-compile(nowarn_deprecated_function).
```

tuncer: I didn't use that because it's too broad and can hide warnings we want to see, but if the pre-18.x warning is a real concern, perhaps we should.

@ferd?

ferd: Would just calling `erlang:now()` get the linter to shut up?

fishcakez: It shouldn't but I think `apply(erlang, now, [])` might.

tuncer: Calling erlang:now() raises the 18.0 deprecation warning, but apply/3 works. Thanks James. See #478.

tuncer: > It shouldn't but I think apply(erlang, now, []) might.

Thanks James, apply/3 works. See #478.

PR #477, opened 2015-04-13
Enable parse transformations in rebar config

kejv:
Only thing which is neccesary is to ensure that parse
transformations are not applied to themselves and probably
also to erl_first_files since they are explicitly stated as
to be compiled before anything else.

### DIFF
--- a/inttest/erlc_dep_graph/src/pascal.erl
+++ b/inttest/erlc_dep_graph/src/pascal.erl
@@ -1,8 +1,8 @@
 %% -*- mode: erlang;erlang-indent-level: 4;indent-tabs-mode: nil -*-
 %% ex: ts=4 sw=4 ft=erlang et
-{erl_opts,
- [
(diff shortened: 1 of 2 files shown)


ferd:
That does sound reasonable.

kejv:
See also original issue: #357

tuncer:
Correct me if I'm wrong, but shouldn't the commit message explain that this is a bugfix, given that the feature is not new?

tuncer:
Other than that, +1.

kejv:
I originally viewed it as a new feature, since I don't see anything in the git history that this use case should ever work. But if you think that it is more a bugfix than a feature, I'm ready to change the commit message appropriately.

tuncer:
Except for some options that are filtered, erl_opts is passed as is to compile:file, so a modified commit message will make a lot more sense.

kejv:
Ok. I have modified the commit message.

PR #482, opened 2015-04-22
Windows runner - CD to node root

carlosedp:
Changes to the Windows runner to CD to the node root directory.

This prevents stuff like lager logging going into the current "log" directory instead of the node release "log" directory.

The unix runner already runs from the node root.


tuncer:
+1

PR #486, opened 2015-04-28
Add eunit_test_failure_ignore option

vojtarylko:
It is pretty common to have such option and quite convenient on
occasion.


tuncer:
Isn't `rebar -k eunit` sufficient?

tuncer:
cc @vladdu

vojtarylko:
Option `-k` also returns non-zero exit code. And it doesn't distinguish between eunit failure and test failure.

vladdu:
My feeling is that options in rebar.config should be for more permanent settings. I would ignore test results only in special cases, when a command line parameter works better. It's too easy to forget the option enabled and think all is well when the tests are failing all over the place.

And as a smaller issue, I think I might want to know that some test have failed, that is I'd like "One or more eunit tests failed." to get printed anyway.

tuncer:
Still pondering about this.

I agree with @vladdu that if this in rebar.config, rebar_eunit better report ignored test results, or else it's easy to forget about. IIRC, ct and ts have the concept of okay-to-fail tests. Doesn't EUnit have something similar?

tuncer:
@richcarl

tuncer:
cc @richcarl

vojtarylko:
@tuncer I don't know about such an option in EUnit.

The information about failing/passing tests is printed out by EUnit. The rebar doesn't have to do it yourself.

tuncer:
> I don't know about such an option in EUnit.

Me neither, but I asked just in case it's hard to find :).

ferd:
I'm wondering whether it wouldn't make more sense to add the option to eunit, rather than having rebar absorb all of the fixes people want for the actual tools. Not that I'm opposed to merging it here, but it appears counter-productive in the long run to encode rules for a test framework across multiple codebases.

vojtarylko:
@ferd I see problem in rebar because of it is interpreting EUnit results.

ferd:
So am I understanding right, from this bit: https://github.com/rebar/rebar/pull/486/files#diff-bfec549e6963d4adb6816dec0113d00bR193 that this PR does not only ignore failures, it entirely ignores test results and reports nothing when there's a failure?

How is this different from, you know, not running the tests? It seems to be it should at the very least still output the message, just not make it an `ABORT`.

tsloughter:
Why run the tests if you don't care that about the result?

vojtarylko:
@ferd @tsloughter We process and interpret EUnit output on our own. This is messed by rebar aborting (so we have to examine reason of such abort).

tsloughter:
It sounds like you want your own eunit plugin to use instead of the one that comes with rebar

vojtarylko:
@tsloughter No, I just want rebar not to abort when everything works well.

tsloughter:
But your definition of "everything works well" appears to be even when tests fail?

vojtarylko:
@tsloughter Yes. EUnit, rebar, erlang - everything works well and I have EUnit output on stderr.

richcarl:
Might be legit to have in rebar, but in eunit, an option to ignore failures
seems quite pointless.


        /Richard

tuncer:
> Might be legit to have in rebar, but in eunit, an option to ignore failures seems quite pointless.

My impression was that @vojtarylko was looking for something like ct's ignore-test-case feature, and that's what I thought would make sense in eunit.

richcarl:
For ignoring individual tests, you mean? Yes, that would be useful.

tuncer:
On Tue, May 5, 2015 at 11:43 AM, Richard Carlsson wrote:
>
> For ignoring individual tests, you mean? Yes, that would be useful.

Yes, individual tests.

vojtarylko:
I'm looking for general "do not abort rebar if some tests fails" so I can distinguish between rebar (or eunit) failing and some tests failing.

tuncer:
On Tue, May 5, 2015 at 1:11 PM, Vojtech Rylko wrote:
>
> I'm looking for general "do not abort rebar if some tests fails" so I can distinguish between rebar (or eunit) failing and some tests failing.

You said that `rebar -k eunit other_cmd` terminating with a non-zero
exit code doesn't work for you.

You also wrote

> And it doesn't distinguish between eunit failure and test failure.

but I'm not sure what "and test failure" refers to. Are you saying
test==one_unit_test and eunit=rebar_eunit?



While we can add something EUnit-specific to rebar, I'm not entirely
sure what the use case is. So let me ask some more questions:

1. You want to run EUnit tests but always ignore the result and check
the logs for success/failure, right?
2. Wouldn't running `rebar eunit` separately from other other rebar
commands solve your use case?
3. Are you sure it's just `rebar eunit`, or how likely are you to
require the same exit(0) `rebar -k` behavior for other commands?
4. How about {keep_going_exit_code, int()} in rebar.config?

tuncer:
On Tue, May 5, 2015 at 1:11 PM, Vojtech Rylko wrote:
>
> I'm looking for general "do not abort rebar if some tests fails" so I can distinguish between rebar (or eunit) failing and some tests failing.

You said that `rebar -k eunit other_cmd` terminating with a non-zero exit code doesn't work for you.

You also wrote

> And it doesn't distinguish between eunit failure and test failure.

but I'm not sure what "and test failure" refers to. Are you saying test==one_unit_test and eunit=rebar_eunit?



While we can add something EUnit-specific to rebar, I'm not entirely sure what the use case is. So let me ask some more questions:

1. You want to run EUnit tests but always ignore the result and check the logs for success/failure, right?

2. Wouldn't running `rebar eunit` separately from other other rebar commands solve your use case?

3. Are you sure it's just `rebar eunit`, or how likely are you to require the same exit(0) `rebar -k` behavior for other commands?

4. How about {keep_going_exit_code, int()} in rebar.config?

vojtarylko:
"eunit failure and test failure" - The eunit test generator can fail and thus not outputting all the tests into the XML. In such case you could have incomplete output of passed tests and a non-zero return code from rebar, and if you ignore the code, you generate all-ok test results, but they're incomplete. The test failure I mean just failing test (eg. ?assert( false )).

ferd:
I think we are not really interested in supporting this at the time. The case where eunit fails to run tests properly to be distinguished from tests running fine but having one fail is minor and no matter which happens, it can be figured out from the output with or without the varying error codes.

Ideally if we were to have a patch for this, I'd say to do something like return 0 when all passes, return 1 when tests fail but ran, and 2 when tests failed and didn't compile right. But no things like "force to return a 0 status when stuff broke."

PR #491, opened 2015-05-08
fix absolute source path on failed build, add gpb proto error test case

lrascao:
Fixes regression introduced in fix PR #489 regarding console output of absolute source paths, expands proto_gpb test case to add coverage for said regression


tuncer:
Other than the minor comment issues, +1. @ferd, this needs to get in quickly.

PR #492, opened 2015-05-09
Start rebar's VM in -mode minimal

tuncer:
This reduces startup time by 100 to 200 milliseconds.


tsloughter:
You don't get a network stack if you do that. Which is why the tests are failing.

tuncer:
I know, distribution is missing. What's the right way to spin it up, or is this impossible after the vm has started?

tsloughter:
I'm pretty sure it is impossible. I dug through the VM code to follow what minimal does at one point but don't remember the details.

tuncer:
I'll close this then. Maybe we can come up with a custom boot file as an alternative.

tuncer:
Here's the the alternative which reduces the overhead even more and doesn't suffer from the same limitation: #493.

PR #493, opened 2015-05-09
Cut startup time

tuncer:


tuncer:
Hmm, actually I should change the default itself in rebar_escripter, as there's no reason not to.

PR #336, opened 2014-08-07
Add details on Dialyzer with "make check"

derek121:
If Dialyzer hasn't been used on an account before, there won't be a PLT
file and "make check" will fail. Added info on that fact to CONTRIBUTING.md.

PR #496, opened 2015-05-12
Update PLT detail for make build_plt

derek121:
The PLT that rebar will have dialyzer use is now created with `make
build_plt` - this commit/PR explains that.

Related to PR https://github.com/rebar/rebar/pull/336


tuncer:
+1

Issue #374, opened 2014-10-23
Windows continuous integration


tuncer:
It'd be great to make use of http://www.appveyor.com for Windows continuous integration.

If the project admins create a rebar account at appveyor and link to it in the repo settings, I can take a look at adding the needed build+test config.

tuncer:
ping?

ferd:
I checked it, and to sync it with github, it wants to have read access to a bunch of stuff in my accounts and organization. It turns out one of my organizations is my workplace, and I'm not willing to open up github permissions to that.

tuncer:
Is creating a `rebar-win-ci` user an option?

ferd:
I know I personally won't have the required permissions to promote it as a maintainer of the org or whatever.

tuncer:
@lrascao can you look into this?

lrascao:
Sure thing, can't use my user as i have the same issue as @ferd.
So i've created the `rebar-win-ci` github user and configured appveyor to build a fork of rebar, works out of the box pretty well, smooth build. I've got the creds of the user with me, any ideas on the best way to share them?
Tests, not so smooth, most file related tests are failing, i'm not on Windows so it's a bit hard to debug, is any of us on Windows?

tuncer:
> Sure thing, can't use my user as i have the same issue as @ferd.
So i've created the rebar-win-ci github user and configured appveyor to build a fork of rebar, works out of the box pretty well, smooth build. I've got the creds of the user with me, any ideas on the best way to share them?

What permissions does the user require for the repo. Does it need write permissions in rebar/rebar to be linked in the hook?

> Tests, not so smooth, most file related tests are failing, i'm not on Windows so it's a bit hard to debug, is any of us on Windows?

I don't know of any current rebar contributor developing on Windows, but there have been past contributors who primarily used Windows. We could try to contact them.

lrascao:
> What permissions does the user require for the repo. Does it need write permissions in rebar/rebar to be linked in the hook?

Two application require permissions:
* AppVeyor:
  * Read org and team membership
  * Access user email addresses (read-only)
*AppVeyor CI:
  * Full control of repository hooks
  * Read org and team membership
  * Access commit status

tuncer:
What's `Full control of repository hooks`?

lrascao:
I imagine it's access to the webhook REST API (https://developer.github.com/v3/repos/hooks/), rebar has Travis so i'm guessing it's already giving out Write Hook access

[...]

PR #498, opened 2015-05-21
Handle all Windows OS types

mhssler:
Avoid matching Osname to work better with Windows 95.

From Erlang/OTP os reference manual:

  os:type() -> {Osfamily, Osname}

  [...]

  In Windows, Osname will be either nt (on Windows NT), or windows (on
  Windows 95). [...] Avoid matching on the Osname atom.

No tests added as I guess the best way to test this is to run all tests on Windows 95.

### DIFF
--- a/src/rebar_deps.erl
+++ b/src/rebar_deps.erl
@@ -139,7 +139,7 @@ setup_env(Config) ->
     {true, DepsDir} = get_deps_dir(Config),
     %% include rebar's DepsDir in ERL_LIBS
     Separator = case os:type() of
-                    {win32, nt} ->
+                    {win32, _} ->
                         ";";
                     _ ->
                         ":"
--- a/src/rebar_port_compiler.erl
+++ b/src/rebar_port_compiler.erl
@@ -365,7 +365,7 @@ port_opt(Config, {env, Env}) ->
 port_opt(_Config, Opt) ->
     Opt.
 
-maybe_switch_extension({win32, nt}, Target) ->
+maybe_switch_extension({win32, _}, Target) ->
     switch_to_dll_or_exe(Target);
 maybe_switch_extension(_OsType, Target) ->
     Target.
--- a/src/rebar_utils.erl
+++ b/src/rebar_utils.erl
@@ -333,7 +333,7 @@ delayed_halt(Code) ->
             halt(Code);
         false ->
             case os:type() of
-                {win32, nt} ->
+                {win32, _} ->
                     timer:sleep(100),
                     halt(Code);
                 _ ->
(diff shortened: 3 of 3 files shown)


ferd:
Any non-windows OS in the win32 family? I guess stuff like ReactOS could technically be there but I wouldn'T expect it to be supported or breaking :/

tuncer:
While this seems okay, and is unlikely to break anything, does Erlang work on Windows 95 at all? The last I remember, it required at least Windows XP (or even newer since a few versions ago). If Erlang works on Windows 95, +1.

@ferd, `os:type/0` on Windows will return one of the following:

* `{win32, windows}`
* `{win32, nt}`
* `{win32, unknown}`

ferd:
I'm fine with the fix I guess. People using windows 95 and trying to compile software from 2015 on it likely have to know they'll struggle with some assumptions made about their 20+ years old OS.

mhssler:
I have not got Windows 95 myself but http://www.erlang.se/productinfo/index.shtml says:

Erlang/OTP is currently available for the following platforms: ...
 * Windows 9x,NT, 2000, 2003, XP

tuncer:
http://www.erlang.se/productinfo/index.shtml also lists VxWorks, so it's clearly outdated.

http://www.erlang.org/doc/installation_guide/INSTALL-WIN32.html:

_We've built on the following platforms: Windows 2003 server, Windows XP Home/Professional, Windows Vista and Windows 7 (32 and 64 bit). You can probably build on Windows 2000, but you will not be able to install the latest Microsoft SDK, so you have to go back to some earlier compiler. Any Windows95'ish platform will surely get you into trouble, what I'm not sure of, but it certainly will..._

That doesn't necessarily mean the result won't run on Windows 95, but I'd expect problems due to missing Windows APIs. Also, Windows 95 isn't supported by Microsoft, and I doubt any Erlang user is a Windows 95 user.

We can merge this, and it won't affect anything, but it won't make Erlang work.

@psyeugenic, am I right in recalling that ERTS won't run on Windows 9x, and that even Windows 2000 is problematic?

ferd:
Hell, XP is no longer actively supported by Microsoft either. I wouldn't feel bad for not supporting an OS that even the vendors of no longer support.

lrascao:
I guess we could this go through Appveyor's tests and see if anything breaks, that is if #374 was closed

tuncer:
Can this be closed?

tuncer:
Should this be closed?

mhssler:
> Should this be closed?

I guess not. I would suggest that you run it through Appveyor's tests and merge it if successful.

Issue #350, opened 2014-08-20
Rebar 2.5.1 crashes with a reltool-generated release on Erlang 17+


choptastic:
It seems that with a reltool generated release, rebar will crash when checking the OTP version due to the version change.  Version 2.5.0 works okay (the commit that breaks it is this: https://github.com/rebar/rebar/commit/873d236ce9b4b88d46c4bb1c6868df70363e8a57)

It crashes on this line: https://github.com/rebar/rebar/blob/master/src/rebar_utils.erl#L391 because the generated release doesn't include the OTP_VERSION file in the releases directory by default.

I'm unsure if this is something that should be fixed within rebar or if it's an oversight by the otp team with the release of Erlang 17, but I thought I'd mention it here.

The full error is as follows:
```erlang
Uncaught error in rebar_core: {'EXIT',
                               {{badmatch,{error,enoent}},
                                [{rebar_utils,otp_release1,1,
                                  [{file,"src/rebar_utils.erl"},{line,391}]},
                                 {rebar_utils,get_arch,0,
                                  [{file,"src/rebar_utils.erl"},{line,92}]},
                                 {rebar_port_compiler,default_env,0,
                                  [{file,"src/rebar_port_compiler.erl"},
                                   {line,574}]},
                                 {rebar_port_compiler,setup_env,2,
                                  [{file,"src/rebar_port_compiler.erl"},
                                   {line,181}]},
                                 {rebar_core,'-setup_envs/2-fun-0-',2,
                                  [{file,"src/rebar_core.erl"},{line,522}]},
                                 {lists,foldl,3,
                                  [{file,"lists.erl"},{line,1261}]},
                                 {rebar_core,maybe_execute,8,
                                  [{file,"src/rebar_core.erl"},{line,293}]},
                                 {rebar_core,process_dir1,7,
                                  [{file,"src/rebar_core.erl"},{line,258}]}]}}
make[3]: *** [get-deps] Error 1
```

I haven't spent any time looking at this too closely to try and fix it yet, I just reverted to 2.5.0 because it works well enough.

Thanks!

senturk:
If anyone still experience this problem: 
please upgrade to the current version (today is 17.4). 
I was running on 17.0 till today, ran into that problem and updated Erlang release 17.4. Now it is working.

PR #499, opened 2015-05-29
Try one more location for the OTP_VERSION file

aronisstav:
If one is building Erlang from sources without making a proper release, there
may be an OTP_VERSION file in the root directory.


aronisstav:
This may also fix some problems reported in #350.

tuncer:
+1, and can you please fix the commit message (superfluous newline after "there") and append your name in THANKS, all as part of the single commit?

aronisstav:
Done! (regarding the commit message I wrapped it before 'there')

PR #501, opened 2015-06-01
Fix otp_release() when OTP_VERSION file not exists

### LINKED ISSUE #350: Rebar 2.5.1 crashes with a reltool-generated release on Erlang 17+

choptastic:
If the OTP_VERSION file doesn't exist at all for (in a reltool-generated-release's for OTP > 17 binaries) we attempt to infer the OTP version from included applications. In this commit, just the erts application is checked but if the erts version is not changed, we could check other applications.

I realize that this PR does not include any tests, I'm more just spitballing here to see if this level of fallback is acceptable. If this is acceptable, then I can spend the extra time to add some tests, but this is more of a conversation stimulator - unless, you actually like this as is, then by all means, merge away!

Alternatively, (and from conversations with @tuncer), a much simpler approach, though one that would sacrifice version granularity, would be to simply return a string like "17.x" if the OTP_VERSION file is not found.

Thoughts?

Fixes #350.

### DIFF
--- a/src/rebar_utils.erl
+++ b/src/rebar_utils.erl
@@ -403,7 +403,15 @@ patch_env(Config, [E | Rest]) ->
 %% ====================================================================
 
 otp_release() ->
-    otp_release1(erlang:system_info(otp_release)).
+    case application:get_env(rebar, cached_otp_release) of
+        undefined ->
+            Vsn = otp_release1(erlang:system_info(otp_release)),
+            application:set_env(rebar, cached_otp_release, Vsn),
+            Vsn;
+        {ok, Vsn} ->
+            Vsn
+   end.
+
 
 %% If OTP <= R16, otp_release is already what we want.
 otp_release1([$R,N|_]=Rel) when is_integer(N) ->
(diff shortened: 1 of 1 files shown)


tuncer:
Isn't it good enough to just use the result of
erlang:system_info(otp_release) and print the no-OTP_VERSION-file warning,
stating the version string we have is probably not specific enough? I don't
like having to maintain a table of erts versions.

tuncer:
Isn't it good enough to just use the result of erlang:system_info(otp_release) and print the no-OTP_VERSION-file warning, stating the version string we have is probably not specific enough? I don't like having to maintain a table of erts versions. Alternatively, as I suggested in #350, we could copy OTP_VERSION into the generated release.

choptastic:
> I don't like having to maintain a table of erts versions.

It also leaves a bit of a sour taste in my mouth, and I do agree with that. It's rather putzy, but gives us some granularity.

As for copying the OTP_VERSION file, I'd be open to that, but for existing releases, that obviously doesn't help. The above would be *okay* in that situation, but not great. I'd also at least like to have it generate a warning about the missing file, and rely upon the erlang:system_info(otp_release) in that case.

Truthfully, as it is right now, I just use 2.5.0 because it just uses erlang:system_info(otp_release)  and doesn't look for the OTP_VERSION file - so that extremely naive version  ("17") is enough for my immediate needs. This was more of a "why not?" kind of a PR - since as I was making my simple fix, I thought, "Hey, this seems like it could be helpful and it's relatively easy to do, even if it has to be updated for every release").

ferd:
Yeah Rebar3 went with just using the major version if the OTP file isn't there and I can't remember if we warn. I forgot to port this back to this version of rebar.

choptastic:
I'll put together a simpler PR for this tonight that doesn't try to infer
the minor version numbers.

PR #502, opened 2015-06-02
Don't crash if missing OTP_VERSION file (fix #350)

### LINKED ISSUE #350: Rebar 2.5.1 crashes with a reltool-generated release on Erlang 17+

choptastic:
This is a simpler version of the fix presented in #501. If generates a WARN if no OTP_VERSION file is found, and resorts to the return of erlang:system_info(otp_release).

I wrapped the otp_release1 function in a simple process dictionary cache to prevent it from being called many times (in my tests it's called quite frequently - like a few hundred times, and since for OTP 17+, there's a hard drive or two for each call, I figured caching the result might speed things up a bit).

Anyway, let me know what you think.

### DIFF
--- a/src/rebar_utils.erl
+++ b/src/rebar_utils.erl
@@ -403,7 +403,17 @@ patch_env(Config, [E | Rest]) ->
 %% ====================================================================
 
 otp_release() ->
-    otp_release1(erlang:system_info(otp_release)).
+    %% We cache the return of otp_release1, since otherwise, we're repeatedly
+    %% reading the same file off the hard drive and generating warnings if they
+    %% aren't there.
+    case erlang:get(otp_release_cache) of
+        undefined ->
+            Vsn = otp_release1(erlang:system_info(otp_release)),
+            erlang:put(otp_release_cache, Vsn),
+            Vsn;
+        Vsn ->
+            Vsn
+    end.
 
 %% If OTP <= R16, otp_release is already what we want.
 otp_release1([$R,N|_]=Rel) when is_integer(N) ->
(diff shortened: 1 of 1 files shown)


tuncer:
+1, even though I'm not sure about the use of pdict here. However, in this special case, like for rebar_log settings, it seems okay to use the process dictionary instead of putting it into rebar_config:config(). Let's get this into 2.6.0, once the minor issues are fixed.

choptastic:
Thanks, wow that was quite the abundance of typos. Yesterday must not have been a great day for me. Typos fixed, rebased, and pushed to the PR.

ferd:
Looks good, got a +1, merging.

choptastic:
Thanks for merging. I do also agree that the pdict leaves a bit of a sour taste in my mouth, at least the way I'm using it here. It gets the job done, but if there are other areas that could benefit from some faux memoization/cache hackery, perhaps there's a value in adding some kind of wrapper function to deal with that. Something like `rebar_utils:cache(Key, Fun)` where `Fun` is only evaluated if `Key` is not found, or whatever.

tuncer:
I'm working on something like that, to be announced soon.

choptastic:
That's great to hear! Thanks for the heads up.

PR #511, opened 2015-06-11
Add and use memoization server

tuncer:


tuncer:
@ferd can we get this merged soon?

PR #513, opened 2015-06-19
Add pr2relnotes tool to generate release notes

jaredmorrow:


tuncer:
Thanks @jaredmorrow. What's the license of the script?

jaredmorrow:
Oh, whatever rebar is under, MIT, BSD, Creative Commons, I don't mind.

tuncer:
Without an explicit license, it'll be Apache 2 licensed then. All source files in rebar are explicitly MIT (Expat) licensed.

I'd prefer an explicit license prologue in the script, if anybody wants to reuse it another project, but you'd have to submit such a patch, so it's up to you.

jaredmorrow:
I will a bit later, thanks.

PR #518, opened 2015-06-25
Generate reserved file for erlc regression test

lrascao:
Reserved files such as ._do_not_compile.erl should be
generated automatically when running the regression
test and not added to the repo. .tar.gz source releases
are not correctly generated when handling such files.


ferd:
Thanks! That sounds good and obviously right and tests pass. Merging.

PR #451, opened 2015-02-09
Cross arch

tonyrog:
Basic cross compilation support.
Environment variables used:
REBAR_TARGET_ARCH set the tool chain name to use.
optional (if gcc fail to determine the word size)
REBAR_TARGET_ARCH_WORDSIZE
optional
REBAR_TARGET_ARCH_VSN if a special version of gcc is requested

PR #520, opened 2015-07-02
Fixed version of #451

tuncer:
This is a fixed version of #451:

 * fix commit messages
 * fix whitespace issues
 * move internal helper functions to correct location
 * unexport internal cross arch helper functions
 * fix 18.0 time API compatibility
 * fix Windows temp dir detection
 * use correct helper function name
 * use sh/2 instead of os:cmd/1
 * match file:delete/1 result
 * use logging macros
 * fix typo
 * fix Dialyzer warnings
 * add Tony Rogvall to THANKS
 * add missing termination clause
 * os type must be win32, not windows
 * match file:write_file/2 result
 * document cross-arch variables
 * simplify env_wordsize/1 (Thanks Fred Hebert)

### DIFF
--- a/src/rebar_utils.erl
+++ b/src/rebar_utils.erl
@@ -98,13 +98,13 @@ is_arch(ArchRegex) ->
             false
     end.
 %%
-%% REBAR_ARCH_TARGET, if used, should be set to the "standard" 
+%% REBAR_TARGET_ARCH, if used, should be set to the "standard" 
 %% target string. That is a prefix for binutils tools.
 %% "x86_64-linux-gnu" or "arm-linux-gnueabi" are good candiates
-%% ${REBAR_ARCH_TARGET}-gcc, ${REBAR_ARCH_TARGET}-ld ...
+%% ${REBAR_TARGET_ARCH}-gcc, ${REBAR_TARGET_ARCH}-ld ...
 %%
 get_arch() ->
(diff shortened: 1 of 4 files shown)


src/rebar_utils.erl
```diff
@@ -673,3 +678,150 @@ cleanup_code_path(OrigPath) ->
...
+env_wordsize(Wordsize) when Wordsize =:= false;
+                            Wordsize =:= "" ->
+    ?WARN("REBAR_TARGET_ARCH_WORDSIZE not set, assuming 32\n", []),
+    "32";
+env_wordsize(Wordsize) ->
+    try list_to_integer(Wordsize) of
+        16 -> "16";
+        32 -> "32";
+        64 -> "64";
+        _ ->
```
ferd: why isn't this just:
```erlang
case Wordsize of
    "16" -> Wordsize;
    "32" -> Wordsize;
    "64" -> Wordsize;
    _ -> ?WARN("REBAR_TARGET_ARCH_WORDSIZE bad value: ~p\n", [Wordsize]), 32
end
```

tuncer: Nice catch, I don't know why @tonyrog checks if it's an integer. Should I simplify the function?

tuncer: Nice catch, I don't know why @tonyrog checks if it's an integer. Thanks, simplified.

PR #522, opened 2015-07-03
Fixed version of #395

tuncer:
* do not use pdict
 * do not enable color support by default. once we have the new
   getopt version, we can add a new type of command line flag
   for that.
 * fix Dialyzer warnings
 * use atom instead of boolean
 * use better name for internal function
 * do not try (and fail) to access rebar's app env in retest test

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -143,3 +143,4 @@ Danil Onishchenko
 Stavros Aronis
 James Fish
 Tony Rogvall
+Andrey Teplyashin
--- a/ebin/rebar.app
+++ b/ebin/rebar.app
@@ -69,7 +69,7 @@
          {log_level, warn},
 
          %% Log colored
-         {log_colored, true},
+         {log_colored, uncolored},
 
          %% any_dir processing modules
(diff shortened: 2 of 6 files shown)


tuncer:
That's a should-never-happen failure. Looking into it.

tuncer:
Fixed

ferd:
Tests fail.

tuncer:
Fixed but not pushed

tuncer:
Pushed

tuncer:
@ferd do you know why the relevant log label regex to match color codes doesn't make a difference (not matched)? If we cannot reliably match that, we might as well simplify pattern and restore the original version. Tried with colored and uncolored rebar.

ferd:
No idea. I remember trying the first time around when this branch was submitted and couldn't figure it out in the time I had to spend on the issue.

tuncer:
Okay, let's merge it, then.

tuncer:
Unless you want to simplify the patterns first.

PR #526, opened 2015-08-03
Automatically clean neotoma-generated erl files

macintux:
Remove `.erl` files on clean that appear to have been converted from `.peg`.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -144,3 +144,4 @@ Stavros Aronis
 James Fish
 Tony Rogvall
 Andrey Teplyashin
+John Daily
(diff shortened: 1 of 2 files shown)


lrascao:
@macintux Would you consider cherry picking the commit from #532 into this PR? It consists of a regression test for the feature you have introduced

tuncer:
What's the status? Is one of the pull requests ready to merge?

lrascao:
I suppose we can merge this one right away and afterwards merge #532 which is the regression test (after i clean it up), ideally both commits should be in the same PR but this is one way of moving forward.

tuncer:
ping?

lrascao:
Closing in favor of #532 which contains this fix plus regression tests

PR #530, opened 2015-08-14
Allow behaviors defined inside xref_extra_path

paulo-ferraz-oliveira:
Use case: xref'ing a module that defined behavior ranch_protocol was
exiting with
{'EXIT',{undef,[{ranch_protocol,behaviour_info, [callbacks],[]}

### DIFF
--- a/inttest/xref_behavior/gen_xref_behavior.erl
+++ b/inttest/xref_behavior/gen_xref_behavior.erl
@@ -1,0 +1,9 @@
+-module(gen_xref_behavior).
+
+-export([behaviour_info/1]). -ignore_xref([{behaviour_info, 1}]).
+
+behaviour_info(callbacks) ->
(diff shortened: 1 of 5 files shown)


ferd:
I'm okay with this change. Could we have a test to go with it? See https://github.com/rebar/rebar/blob/master/CONTRIBUTING.md

paulo-ferraz-oliveira:
Thanks for your feedback, Fred. I'll try to have a test case ready soon (not this weekend though) that illustrates the desired behavior (should I have one for success and another for non-success?) and then ammend my commit.

ferd:
Success alone should be good enough.

paulo-ferraz-oliveira:
I updated the pull request. Created only one test case for retest, which illustrates the expected behavior.

PR #537, opened 2015-08-24
Fix eunit test on OTP18

lrascao:
OTP18 changed the output failed assertion string
from assertEqual_failed to assertEqual

### DIFF
--- a/test/rebar_eunit_tests.erl
+++ b/test/rebar_eunit_tests.erl
@@ -131,7 +131,7 @@ eunit_with_suites_and_tests_test_() ->
 
                {"Selected suite's generator test raises an error",
                 ?_assert(string:str(RebarOut,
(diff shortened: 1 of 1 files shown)


ferd:
If I get the tests right, this works in old versions because we match on a partial string and `assertEqual` still matches `assertEqual_failed`? This does seem a bit brittle (it would match on `assertEqual_success`) but fine by me.

lrascao:
luckily there's no `assertEqual_success` but i see your point, the actual string that is being returned is:
**error:{assertEqual_failed,[{module,myapp_mymod3},
                     {line,8},
                     {expression,"false"},
                     {expected,true},
                     {value,false}]}

ideally we should be able to parse the whole string into an erlang term and match that, much safer that way

ferd:
Agreed. for the time being this is fine, though.

Issue #182, opened 2013-12-01
Adapt built-in port_env to support OS X Mavericks or newer


tuncer:
Someone with access to OS X Mavericks or newer should adapt the built-in [port_env](https://github.com/rebar/rebar/blob/988c9cd574538d279193f548297d80ad3b0a45d0/src/rebar_port_compiler.erl#L579).

cc @jaredmorrow @Vagabond @dizzyd

lrascao:
Fixed in #590

PR #538, opened 2015-08-24
Fix #536

### LINKED ISSUE #536: -m64 CFLAG set although unavailable
Rebar sets the -m64 CFLAG on architectures where it is not available, e.g. Debian's arm64 (aarch64-linux-gnu) or alpha (alpha-linux-gnu).
This causes the build process to fail.

See for example
https://buildd.debian.org/status/package.php?p=erlang-sqlite3
https://bugs.debian.org/795932

Maybe instead of hard-coding it for every possible architecture, rebar could probe the compiler?

tuncer:


ferd:
Doesn't this also remove the flags from linux environments where it *is* available and could be relied on to build?

tuncer:
Yes, it does. This was introduced in 2.6.0 (6fca763c9bb70d370d6782c436d5fca286db77df):
```
Fix compiler invocation on multiarch Linux

This fixes port compilation for some configurations, notably this one:
Debian with 32-bit userspace, 64-bit kernel and 64-bit erlang packages.
```

Because the regex is too broad, it causes build errors. I don't know if we can fix it in the regex instead (and restore `-m64` in LDFLAGS). What I do know is that 2.6.0 causes build errors with some toolchains.

/cc @liskin.

Slightly off-topic but related:
@ferd, I don't have access to OS X, but I wonder why #182 isn't an issue. I suppose the toolchain doesn't require extending it for newer versions anymore, as I doubt projects fix this in rebar.config for all OS X versions.

tuncer:
So, what do we do?

liskin:
As I think I said elsewhere, I haven't used rebar (and Erlang) for months so I don't really care very much. If my changes cause trouble for people, I think it's fair that I just put those settings in my own rebar config.

tuncer:
@fred?

PR #539, opened 2015-08-26
Add erl_first_files in eunit_first_files and qc_first_files

platinumthinker:
Fix compile order for tests

### DIFF
--- a/src/rebar_erlc_compiler.erl
+++ b/src/rebar_erlc_compiler.erl
@@ -236,8 +236,10 @@ test_compile_config_and_opts(Config, ErlOpts, Cmd) ->
     Opts = [O || O <- Opts0, O =/= no_debug_info],
     Config4 = rebar_config:set(Config3, erl_opts, Opts),
 
+    FirstErlFiles = rebar_config:get_list(Config4, erl_first_files, []),
     FirstFilesAtom = list_to_atom(Cmd ++ "_first_files"),
-    FirstErls = rebar_config:get_list(Config4, FirstFilesAtom, []),
+    FirstErlsOther = rebar_config:get_list(Config4, FirstFilesAtom, []),
+    FirstErls = FirstErlFiles ++ FirstErlsOther,
     Config5 = rebar_config:set(Config4, erl_first_files, FirstErls),
     {Config5, Opts}.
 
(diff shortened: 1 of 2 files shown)


tuncer:
This looks reasonable and needs a test before merge, but why did you restrict it to "qc" and "eunit"?

platinumthinker:
I saw this behaviour in tests. I add erl_first_files in all first_files..
I think about general first_files instead of separate.

tuncer:
You can skip the list_to_atom call and just use `erl_first_files` in the new version.

Regarding tests, it makes sense to check the new behavior explicitly in a test, maybe as follows:

1. eunit with first rebar.config, which does not have erl_first_files, and expect build failure
2. clean
3. eunit with second rebar.config, which has erl_first_files, and expect build+eunit success

tuncer:
ping?

platinumthinker:
I have been on vacation. I am back now and continue to work on this.

platinumthinker:
ping?

tuncer:
1. The unneeded `list_to_atom("erl_first_files")` is still there.
2. You're not testing the case where the test fails without erl_first_files. Thinking about it, I don't think you can test, because erlc_compiler detects a parse_transform compiler directive and automatically adds it as a dependency. I'm not sure how to test the failure case.

tuncer:
1. The unneeded `list_to_atom("erl_first_files")` is still there.
2. You're not testing the case where the test fails without erl_first_files. However, I don't think you can test, because erlc_compiler detects a parse_transform compiler directive and automatically adds it as a dependency. I'm not sure how to test the failure case.

tuncer:
1. The unneeded `list_to_atom("erl_first_files")` is still there.
2. You're not testing the case where the test fails without erl_first_files. However, I don't think you can test that, because erlc_compiler detects a parse_transform compiler directive and automatically adds it as a dependency. I'm not sure how to test the failure case.

platinumthinker:
I agree with you on second position. First thing is fixed.

tuncer:
You should just use the atom `erl_first_files`. No variable needed.

tuncer:
You should simply use the atom `erl_first_files`. No variable needed.

tuncer:
There's no need to quote erl_first_files.

platinumthinker:
This true, but not mandatory.
Example in this file:
```41: -type erlc_info_v() :: {digraph:vertex(), term()} | 'false'.```
```210:                     {platform_define, "R13", 'old_inets'}]},```

tuncer:
Yes, but (1) in type specs it's customary to quote atoms and (2) we don't quote atoms in any other rebar_config call, which makes this case break the consistency for no reason.

platinumthinker:
okay

tuncer:
Thanks. +1, once you've squashed the fixups into the original commit.

platinumthinker:
Squashed is done

PR #545, opened 2015-09-17
Fix #544

### LINKED ISSUE #544: `whoami` should not be used in priv/templates/simplenode.runner - use `id -un` instead.
The `whoami` command, used in the start/stop scripts, has been obsoleted (`man whoami` on my Mac, and some other OSs, states that), and should not be used. The command is not available anymore on SmartOS and Solaris 11. It should be replaced by `id -un`.

For compatibility with older systems, perhaps the script could try `whoami`, and if it fails, try `id -un`.

ccrusius:
'whoami' is not POSIX-compliant, but 'id -un' is. This is true
at least since the 1003.1-2001 standard, I don't have a copy of the
previous 1003.2-1992 to verify it is true there too.

This causes problems in recent operating systems where 'whoami' is
not even shipped anymore.

### DIFF
--- a/priv/templates/simplenode.runner
+++ b/priv/templates/simplenode.runner
@@ -6,6 +6,10 @@
 if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
     POSIX_SHELL="true"
     export POSIX_SHELL
+    # To support 'whoami' on old Solaris systems, add /usr/ucb to path
+    # (New Solaris systems do not have 'whoami')
+    PATH=/usr/ucb:$PATH
+    export PATH
     exec /usr/bin/ksh $0 "$@"
 fi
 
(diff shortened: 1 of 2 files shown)


tuncer:
Thanks, commented in #544.

tuncer:
You forgot to restore the SunOS ucb PATH extension. Other than that, I agree with the patch. Thanks again.

ccrusius:
The `ucb` path should not be necessary now that `id -un` is used. There's no harm in adding it again, though.

tuncer:
Doesn't the whoami fallback require the ucb path?

tuncer:
Don't forget to append your name in THANKS.

ccrusius:
The `whoami` fallback only requires the `ucb` path on SunOS systems that do not have `id -un`, and I don't know of any. Does not hurt having it there, so I put it back.

tuncer:
+1

tuncer:
@ferd?

PR #546, opened 2015-09-26
Added report and verbose options for LFE compile.

oubiwann:
Recent builds of LFE have stopped reporting compile errors when compiled with ``rebar``. Examination of this issue revealed that this was due to recent changes in default LFE compile options.

The way that LFE handles compile options has slightly changed; ``verbose`` and ``report`` are now provided as defaults only if no other options are passed; if other options are passed, these do not get set. As a result, we have stopped seeing compile error info when building with ``rebar``.

This change brings back the previous level of reporting that LFE developers had when compiling their projects with ``rebar``.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -144,3 +144,4 @@ Stavros Aronis
 James Fish
 Tony Rogvall
 Andrey Teplyashin
+Duncan McGreggor
--- a/src/rebar_lfe_compiler.erl
+++ b/src/rebar_lfe_compiler.erl
@@ -71,7 +71,9 @@ compile_lfe(Source, _Target, Config) ->
             ?FAIL;
         _ ->
             ErlOpts = rebar_utils:erl_opts(Config),
-            Opts = [{i, "include"}, {outdir, "ebin"}, return] ++ ErlOpts,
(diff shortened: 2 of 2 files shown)


tuncer:
+1

tuncer:
Thanks. +1, once you've:

1. appended your name in THANKS
2. removed the trailing period in the commit message title/summary
3. `s/Added/Add/` in commit title/summary
4. added the description as the commit message body

oubiwann:
Done!

rvirding:
This is actually very close to how the erlang compiler handles options, though it does allow prepending a **single** option to the default options which contain verbose. However giving a list of options completely replaces the default options.

tuncer:
@ferd?

oubiwann:
@tuncer I think that @rvirding was simply providing supplemental information, giving more details on the sensible way in which LFE handles its compile options ...

rvirding:
Yes, I was just pointing out that the LFE compiler handles it's options
list in the same way the Erlang compiler does. The return values are now a
little different in details as the new compiler can handle files containing
multiple modules which can produce multiple beam files.

Robert

tuncer:
@oubiwann I've pinged @ferd so that he can merge (or reject). Sorry, I should have written that explicitly.

PR #552, opened 2015-10-16
Proto compiler needs an protos_pb.erl src_out_dir option

dcarastan:
protos_pb.erl comes handy when one tries to parse a stack trace. This pull request capture a possible implementation for this issue. While working on it I also noted a port compilation issue that came from compiler's inability to handle the extra quotes. On a Mac:

c_src/bcrypt_port.c:20:10: fatal error: 'erl_interface.h' file not found
# include <erl_interface.h>

```
     ^
```

1 error generated.

P.S. Guidelines noted too late. I will try to re-work this when I get a break. Will definitely keep them in mind. Thanks!


tuncer:
Even though I don't think we can apply it as is, because the quoting was added to fix other issues, can you split the port_compiler patch into a separate pull request?

lrascao:
Besides the point @tuncer made about moving the port_compiler to a separate pull request, for consistency, maybe we should define the `src_out_dir`option in a `protobuffs_opts` tuple in much the same way as is being done with `gpb_opts`, ideally with the same name (gpb uses `o_erl`and `o_hrl` for .erl and .hrl).
The checks are failing due to a xref error, could you please run `make xref`and fix the issues?

lrascao:
@dcarastan ping?

dcarastan:
@lrascao: Just started work at another company. It is very difficult to find the time for it and won't be able to re-work the pull request. Shall I cancel it?

lrascao:
That's ok, it's a valuable contribution, just leave it as is in case you find the time to complete it.
thanks!

tuncer:
bump?

tuncer:
I think this can be closed due to inactivity.

Issue #550, opened 2015-10-13
Templates cannot cope with whitespace in paths


evnu:
Some scripts in `priv/templates` have problems with paths containing whitespace. For example, `simplenode.runner`:

```
[mmueller@entelios templates]$ pwd
/tmp/rebar whitespace/priv/templates
[mmueller@entelios templates]$ ./simplenode.runner start
cat: /tmp/rebar: No such file or directory
cat: whitespace/priv/releases/start_erl.data: No such file or directory
grep: /tmp/rebar: Is a directory
grep: whitespace/priv/etc/vm.args: No such file or directory
vm.args needs to have either -name or -sname parameter.
```

I identified `simplenode.runner` and `simplenode.nodetool` to show problems like this. Maybe other scripts are affected as well.

evnu:
To find missing quotation of variables in shell scripts, [shellcheck](https://github.com/koalaman/shellcheck) can be used. shellcheck returns a lengthy list for `simplenode.runner`. Example:

    In simplenode.runner line 6:
    if [ `uname -s` = 'SunOS' -a "${POSIX_SHELL}" != "true" ]; then
         ^-- SC2046: Quote this to prevent word splitting.
         ^-- SC2006: Use $(..) instead of deprecated `..`
    In simplenode.runner line 13:
        exec /usr/bin/ksh $0 "$@"
                                       ^-- SC2086: Double quote to prevent globbing and word splitting.
    In simplenode.runner line 19:

     ......snip.....

tuncer:
@evnu, would you like to submit a patch?

evnu:
@tuncer I will give it a try.

Note that OTP itself has problems with whitespace in paths: https://github.com/erlang/otp/pull/849. Thus, patching the issues here will move the problem further upstream.

tuncer:
Thanks.

rebar has to work with all kinds of OTP versions, including those that haven't fixed the issue. Therefore, it's wise to fix what we can in rebar and the shell scripts.

evnu:
The changes to fix this issue will be rather large. I would like to add unit tests to check that this does not break the wanted behaviour. What is the right place to add such tests?

evnu:
Regarding my note on erlang/otp#849: when using the arguments which do not invoke `escript` (such as `simplenode.runner console`), the referenced bug does not apply, as `erl` is called directly.

tuncer:
> What is the right place to add such tests?

`inttest/`.

ccrusius:
[...]

PR #554, opened 2015-10-19
Add integration tests for simplenode.runner

evnu:
This changeset adds integration test for `priv/templates/simplenode.runner`. It tests the script against an example release and also tries to perform a release upgrade.

Tested on ArchLinux and Apple Mac OS X 10.10.4..


evnu:
The timeouts in the patchset are not very large (1s mostly). This could be a problem for travis.

lrascao:
Maybe we should have a VW component that increases these timeouts for Travis builds

evnu:
For OTP versions < 17, the testsetup is faulty. I will fix that now.

evnu:
I extended the test suite to also cover OTP versions before R15B01. Running the tests succeeds locally, but it seems that I am still having problems with travis.

evnu:
I fixed the problems with older Erlang releases (https://github.com/rebar/rebar/commit/26343299ec7e2c77757651590d7548832adbf4f0). Unfortunately, I did not find a way to fix some test cases in `basic_rt.erl` for Travis. I think I am missing something about how to use the `-heart` option with Travis. Also, `start_boot` does not start a node properly, but I could not extract any log files or error messages for that.

evnu:
I fixed the remaining problems with travis in https://github.com/rebar/rebar/commit/ca30203f7b0579ef0720268a7c0a250f2b4ec789. I removed certain test cases (`start_boot`, `restart`, `reboot`), as I could not get them to run reliably with travis.

evnu:
With https://github.com/evnu/rebar/commit/2f7880d8416716af4c74e361ee1c66ecc85012a5, I removed the trailing . from the first commit summary as asked for by @tuncer  in #550.

tuncer:
Bump?

PR #557, opened 2015-10-24
Fix windows eunit tests

lrascao:
File tests: windows file operations should abort on error
the same as unix operations invoked through sh. Also
windows does not support '?' characters in filenames.

Eunit tests: the 'All' prefix is missing on the 'x tests passed'
message. Eunit only prints the 'All' prefix if there are more
than 2 passed tests, dropping the prefix on the match works
for all cases.

### DIFF
--- a/src/rebar_file_utils.erl
+++ b/src/rebar_file_utils.erl
@@ -88,7 +88,7 @@ mv(Source, Dest) ->
                         ?FMT("move /y \"~s\" \"~s\" 1> nul",
                              [filename:nativename(Source),
                               filename:nativename(Dest)]),
-                        [{use_stdout, false}, return_on_error]),
+                        [{use_stdout, false}, abort_on_error]),
             case R of
                 [] ->
                     ok;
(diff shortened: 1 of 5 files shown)


ferd:
That looks good, +1.

tuncer:
1. Why do we need `rebar.bat` in addition to the already generated `rebar.cmd`? Does it behave differently?
2. Are you sure we weren't explicitly testing filenames with spaces? If you aren't, shouldn't we try to quote it instead of removing the spaces?

Other than that, +1.

tuncer:
1. Why do we need `rebar.bat` in addition to the already generated `rebar.cmd`? Does it behave differently?
2. Are you sure we weren't explicitly testing filenames with spaces? If you aren't, shouldn't we try to fix the incompatibility instead of removing the spaces?

Other than that, +1.

lrascao:
1. right you are, no need for .bat whatsoever, didn't notice the .cmd being generated by bootstrap
2. after further investigation, turns out the only thing that was causing the test to fail was the `?`character which windows does not allow to be part of directory names, spaces are fine

tuncer:
+1

PR #559, opened 2015-10-27
Fix paths for libraries and headers from OTP 18.1 and add OTP 18.1 to the list of test environments.

pprzetacznik:
OTP 18.1 has different paths for erts headers and erl_interface libraries. This pull request not only adds paths for files in new locations but also keeps old paths to provide support for older versions of OTP.

In second commit I add otp 18.1 to the list of test environments.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ otp_release:
   - R14B04
   - R14B03
   - 17.0
+  - 18.1
 before_script:
   - hostname -f
   - cc -v
--- a/src/rebar_port_compiler.erl
+++ b/src/rebar_port_compiler.erl
@@ -524,8 +524,19 @@ erts_dir_short() ->
 erts_dir() ->
     lists:concat([erts_dir_short(), "-", erlang:system_info(version)]).
 
-otp18_target() ->
+otp_src_target() ->
     os:cmd("$ERL_TOP/erts/autoconf/config.guess | tr -d '\n'").
+
+is_subdir(_Dir, false) ->
+    install;
+is_subdir(Dir, Subdir) ->
+    case lists:prefix(Subdir, Dir) of
+        true -> src;
+        false -> install
+    end.
+
+otp_type() ->
+    is_subdir(os:find_executable("erl"), os:getenv("ERL_TOP")).
 
 os_env() ->
     ReOpts = [{return, list}, {parts, 2}, unicode],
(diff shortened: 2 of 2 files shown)


tuncer:
Thanks.

1. This must have gone unnoticed because there's no erl_interface test case. We should add one as part of this pull request.
2. `otp18_target/0` references a file from the OTP build tree. We cannot rely on that here.
3. Why do we need both `ERL_EI_LIBDIR2` and `ERL_EI_LIBDIR`?
4. Why don't we simply reference `usr/lib/erlang/usr/lib` and `usr/lib/erlang/usr/include` instead? These exist in R13B03 and 18.1.3. But maybe I'm missing the incompatible change.

pprzetacznik:
2. Do you have any proposition for improving that?
3. `ERL_EI_LIBDIR` refers to specific directory. I left that because it may be used somewhere. Another option is to remove both `ERL_EI_LIBDIR` and `ERL_EI_LIBDIR2` and try to assemble it like it was done with `ERL_CFLAGS`
4. These modifications refer to directory structure in otp compiled from sources, eg. `otp_src_18.1`. Probably `rpm` packages or `make install` move these files to the directories that are previously included. I've just focused on otp 18 compiled from source package and linked under `$ERL_TOP` variable.

tuncer:
Okay, it wasn't clear from the description that these are issues one faces when using OTP that's been built but not installed. In that context `$ERL_TOP` makes sense. The cleanest approach would be to figure out that we are using not-yet-installed-just-built-OTP and switch to a different code path, which should leave the usual code path as it was before. Maybe the presence of `$ERL_TOP` plus `rebar_utils:find_executable("erl")` being inside that directory is enough as a check. This of course means `otp18_target/0` can rely on `$ERL_TOP`, but the function should be renamed, as this is hardly specific to 18.x. Maybe name it `otp_src_target/0` and use the same scheme for similar functions we have to introduce in the proposed if-otp_src refactoring.

pprzetacznik:
@tuncer I've extracted `is_build_otp()` function and reduced removed `ERL_EI_LIBDIR2` variable by using pattern matching if I'm using built otp. Would you mind to check if this code is ok right now?

tuncer:
Other than the two code suggestions, the changes look good now.

Do you think you can add a test exercising erl_interface (in `inttest/`)? That would be great.

pprzetacznik:
I've applied your suggestions and amended previous commit. Unfortunately this removed your two great comments. Would you like to take a look now?

And about tests for `erl_interface`. Unfortunately due to my lack of knowledge about project structure I see them hard to implement. Probably you would have to provide me with more specifics.

tuncer:
> I've applied your suggestions and amended previous commit. Unfortunately this removed your two great comments.

Incomplete history is a feature of GitHub's code review system. If you had added another fixup commit, I would have been able to review just the fixes, and you could have made a final squash before the merge. I'm glad you did add an extra commit to fix the issues from the first commit in the branch. That made it much easier to review only the fixes. For a better code review system look at Gerrit.

> Would you like to take a look now?

Looks good.

> And about tests for erl_interface. Unfortunately due to my lack of knowledge about project structure I see them hard to implement. Probably you would have to provide me with more specifics.

Have a look at `inttest/port/port_rt.erl`. It should be easy to add another C file and extend `port_rt.erl`.

```
# fetch+build retest
$ make deps
# run port compiler tests
$ deps/retest/retest -v inttest/port
```

The process should be documented in CONTRIBUTING.md. If it isn't (clear enough), I'd welcome a patch.

tuncer:
> I've applied your suggestions and amended previous commit. Unfortunately this removed your two great comments.

History may get lost in GitHub's code review system. If you had added another fixup commit, I would have been able to review just the fixes, and you could have made a final squash before the merge. I'm glad you did add an extra commit to fix the issues from the first commit in the branch. That made it much easier to review only the fixes.

> Would you like to take a look now?

Looks good.

> And about tests for erl_interface. Unfortunately due to my lack of knowledge about project structure I see them hard to implement. Probably you would have to provide me with more specifics.

Have a look at `inttest/port/port_rt.erl`. It should be easy to add another C file and extend `port_rt.erl`.

```
# fetch+build retest
$ make deps
# run port compiler tests
$ deps/retest/retest -v inttest/port
```

The process should be documented in CONTRIBUTING.md. If it isn't (clear enough), I'd welcome a patch.

tuncer:
@pprzetacznik, bump?

PR #566, opened 2015-11-12
Allow Windows release to get vm.args and sys.config from node_root/et…

ssbx:
Add support for looking at %node_root%\etc to find vm.args and sys.config if not found in %release_dir%\%release_version%. This is a copy/paste from simplenode.windows.start_erl.cmd that already implement it.


lrascao:
@ssbx would you mind squashing your commits into a single one? Please keep in mind the commit guidelines in CONTRIBUTING.md:
`
One line summary (at most 50 characters)

Longer description (wrap at 72 characters)
`

Issue #572, opened 2015-12-08
rebar generate-upgrade


WanliTian:
I have one problem when using rebar to generate upgrade packet.

My application have some dependencies  including hackney.The hackney applicaiton has one **included_applications** called `ssl_verify_hostname`.

Now i use `rebar generate-upgrade previous_release=test-pre` to genereate my upgrade packet, but i got `ERROR: Systools [systools:make_relup/4] aborted with: [{error_reading,
                                                        {ssl_verify_hostname,
                                                         {not_found,
                                                          "ssl_verify_hostname.app"}}}]`

How can i fix this problem?

btw:
Is there are some plants for automatically generating appups about changing supervisor's workers

lrascao:
Do you have a simple project where this issue occurs? also can you please send the output of `rebar -v generate-upgrade previous_release=test-pre`

WanliTian:
try this:

1. `mkdir hotrep && cd hotrep`
2. create a project with `rebar create template=simpleapp appid=hotrep`
3. add an dependent application for rebar.config(**hackney**) and change hotrep.app.src
4. generate a release 
5. change this application's version
6. generate new release , appup and upgrade packet.

when using `rebar` to generate upgrade packet, you will get the error that i got

WanliTian:
using this -v option ,i got this:

`ERROR: Systools [systools:make_relup/4] aborted with: [{error_reading,
                                                        {ssl_verify_hostname,
                                                         {not_found,
                                                          "ssl_verify_hostname.app"}}}]
ERROR: 'generate-upgrade' failed while processing /home/users/WanliTian/Code/Erlang/hotrep/rel: rebar_abort`

lrascao:
also, could you please check your rebar version? `rebar --version`

WanliTian:
I will try it later
thx for your reply.

my rebar version is `rebar 2.6.0 18 20150910_080108 git 2.6.0-24-g2941760`

WanliTian:
i rebuild rebar with the latest version`rebar 2.6.1 18 20151211_090405 git 2.6.1-16-gb19a839`, and i get the same result.

`ERROR: Systools [systools:make_relup/4] aborted with: [{error_reading,
                                                        {ssl_verify_hostname,
                                                         {not_found,
                                                          "ssl_verify_hostname.app"}}}]
ERROR: 'generate-upgrade' failed while processing /home/users/WanliTian/Code/Erlang/hotrep/rel: rebar_abort`

[...]

PR #573, opened 2015-12-13
Provide additional debug logging on relup generation

lrascao:
Issue #572 exposed the need for better logging information


tuncer:
+1

PR #578, opened 2016-01-15
Add the latest OTP 18 version to the Travis build

lrascao:


tuncer:
See also #295. Once kerl is fixed, travis-ci will be able to make all otp releases available.

lrascao:
Cool, hadn't noticed that one, do you want to try and put #295 through Travis again and see how it behaves?

tuncer:
Sure, but not before next week. In the meantime, feel free to patch kerl :).

tuncer:
To be clear, the missing otp versions won't magically appear. kerl needs to be fixed to build from git tag tarballs, not just what's inside erlang.org/download.

tuncer:
Updated #295.

PR #585, opened 2016-02-04
Fix FQDN hostname check for unix

posilva:
I have generated a release using "rebar generate" and deploy it to a AWS EC2 instance. 
After running "/srv/erlapp/erlapp start" the following error appear: 

"Hostname must be a fqdn domain name when node is configured with long names
and the full node name isn't configured in vm.args"

Running the 'hostname' command i get: ip-10-2x-xxx-x and running 'hostname -f' is 10-2x-xxx-x.us-west-2.compute.internal 

After update the 'erlapp' script to use 'hostname -f' the node started correctly.

### DIFF
--- a/priv/templates/simplenode.runner
+++ b/priv/templates/simplenode.runner
@@ -103,7 +103,7 @@ REMSH_NAME=`echo $NAME_ARG | awk '{print $2}'`
 
 # Test if REMSH_NAME contains a @ and set REMSH_HOSTNAME_PART 
 # and REMSH_NAME_PART according REMSH_TYPE
-MAYBE_FQDN_HOSTNAME=`hostname`
+MAYBE_FQDN_HOSTNAME=`hostname -f`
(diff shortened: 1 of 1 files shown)


lrascao:
The `hostname` check was introduced in #115, the `-f` flag was discussed there and deemed to be risky since it's not portable.
relx has an elegant solution for this, it invokes a escript that prints out the `node()` method, this seems much more portable than the `hostname`call

tuncer:
Yeah, so why not change the line to something like this?

```
MAYBE_FQDN_HOSTNAME=`erl -name foo -eval -noinput -eval 'io:format("~s~n", [node()]), halt(0).'`
```

tuncer:
Yes, so why not change the line to something like this?

```
MAYBE_FQDN_HOSTNAME=`erl -noinput -eval 'io:format("~s~n", [node()]), halt(0).'`
```

tuncer:
bump?

PR #586, opened 2016-02-12
Ensure ebin created for dia compiler

surik:
Previously if `ebin` was not created beams for compiled did files were not stored

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -147,3 +147,4 @@ Andrey Teplyashin
 Duncan McGreggor
 Sebastien Serre
 John Daily
+Yury Gargay
--- a/src/rebar_dia_compiler.erl
+++ b/src/rebar_dia_compiler.erl
@@ -88,6 +88,8 @@ info_help(Description) ->
 compile_dia(Source, Target, Config) ->
     ok = filelib:ensure_dir(Target),
     ok = filelib:ensure_dir(filename:join("include", "dummy.hrl")),
+    ok = filelib:ensure_dir(filename:join("ebin", "dummy.beam")),
+    true = code:add_path(filename:absname("ebin")),
     Opts = [{outdir, "src"}] ++ rebar_config:get(Config, dia_opts, []),
     case diameter_dict_util:parse({path, Source}, []) of
         {ok, Spec} ->
--- a/test/rebar_dia_tests.erl
+++ b/test/rebar_dia_tests.erl
@@ -1,0 +1,116 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil -*-
+%% ex: ts=4 sw=4 et
+-module(rebar_dia_tests).
+
+-compile([export_all]).
+
+-include_lib("eunit/include/eunit.hrl").
+
+-define(REBAR_SCRIPT, "../rebar").
+
+-define(TMP_DIR, "dia_project/").
+
+dia_test_() ->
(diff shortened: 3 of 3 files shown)


lrascao:
@surik, change looks sane, i've just now noticed we've got no test coverage on the dia compiler.
Would you be available to write a small test to improve rebar coverage? If not, then real world .dia files would also be helpful.

surik:
I added a small test for dia compiler and this test shows that compiler doesn't work on OTP 14 because uses undefined function `diameter_dict_util:parse/2`

lrascao:
I see, maybe that explains why there were no tests.
We can check for versions older than R15 and ignore the test for those. `rebar_reltool.erl` already does this (see `check_vsn/0`).
Also should we not be checking for correct output from the dia compiler? A simple comparison with a static string of the expected output would suffice.
For both checking the OTP version and the expected output maybe Eunit tests are a bit limited, integration tests (in the `inttest/` folder) perhaps are more suitable for this case, you can find a similar test  in `inttest/proto_gpb`, it takes a `.proto` file and compiles it to `.erl` while checking for correct output.
I understand that this has a bit more scope than you intended with the original PR, please let me know if you need a hand with this, i'll be glad to collaborate with you and take this chance to improve rebar's support of diameter compiling.

surik:
I added ignoring test for dia compiler on OTP < 15.

Tests on R16B02 are broken and looks this is not related to my changes.

lrascao:
> Tests on R16B02 are broken and looks this is not related to my changes.
They were not, there's one specific test (`ct1_rt`) that needs a larger timeout

> Also should we not be checking for correct output from the dia compiler? A simple comparison with a static string of the expected output would suffice.
> For both checking the OTP version and the expected output maybe Eunit tests are a bit limited, integration tests (in the inttest/ folder) perhaps are more suitable for this case, you can find a similar test in inttest/proto_gpb, it takes a .proto file and compiles it to .erl while checking for correct output.

What are your thoughts on this subject?

surik:
> What are your thoughts on this subject?

I fixed a small issue and provided a test which shows that it is works well. EUnit is enough in this case. What would be a different with `inttest`?

Maybe we can write a better test for dia compiler in `inttest` but in my opinion it is out of scope of this PR.

lrascao:
> I fixed a small issue and provided a test which shows that it is works well. EUnit is enough in this case. What would be a different with inttest?

The test shows that something was compiled, a complete test would show that the output format is as expected. Eunit is apropriate for unit testing, `inttest` allows for more complex checks (integrity, making sure that the file is not compiled more than once unnecessarily, etc).

> Maybe we can write a better test for dia compiler in inttest but in my opinion it is out of scope of this PR.

I understand, that's why i offered my help in making this a more complete PR that contains tests. If this is not possible i'd prefer this PR with just the fix and i'll propose a separate PR with more thorough testing.
What are your thoughts on this @tuncer?

tuncer:
We can extend the test and fix R16 later, no need to block the pr.

lrascao:
+1 then, maybe squash 99d1c5a2431afe8782e8691b8812712fb9215831 and 60861a44b56bbdc0800a8ad46613d8c1bd08a197 first?

surik:
Squashed

lrascao:
Thanks, if you wish please add your name to the THANKS file if not already there

PR #587, opened 2016-02-14
rebar_require_vsn: Update regexp to account for newer OTP versions

lrascao:
Add support for patch numbers,  also expose
version_tuple/1 for easier internal access.
Update regression tests accordingly.


ferd:
+1

PR #588, opened 2016-02-15
Look for ct .spec files in the ct_dir that was specified

lrascao:
Change existing behaviour which is to find all .spec
files recursively in the current working directory.
This is confusing since the user explicitly stated the
location for his spec files and negates the possibility
to have different spec'ed suites for different environment.

This is a potentially breaking change, projects that have several .spec files across different dirs could be counting on this behaviour, they would have to group them under a single dir.
However this change allows projects to define (using the rebar.config.script facility) separate .spec files with different test suites, so for example under Travis you could run a lighter set of tests and your machine run the full test battery, currently you have no way to accomplish this (with test .spec that is)

### DIFF
--- a/inttest/ct4/ct4_rt.erl
+++ b/inttest/ct4/ct4_rt.erl
@@ -1,0 +1,44 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil -*-
+%% ex: ts=4 sw=4 et
+-module(ct4_rt).
+
+-compile(export_all).
+
+setup([Target]) ->
+  retest_utils:load_module(filename:join(Target, "inttest_utils.erl")),
+  ok.
+
+files() ->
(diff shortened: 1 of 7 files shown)


lrascao:
@tuncer what are your thoughts on this?

tuncer:
On 12 March 2016 at 12:12, Luis Rascao wrote:
>
> @tuncer what are your thoughts on this?

I'm concerned that this will break existing projects. Can we detect
the scenario and either (1) warn and use new behavior or (2) warn and
suggest a new opt-in flag?

lrascao:
i like (2), let's go with that

tuncer:
Did you forget to document it in rebar_ct:info?

lrascao:
I did, fixed now

PR #589, opened 2016-02-18
Introduce REBAR_DEPS_PREFER_LIBS env variable to alter search behaviour

matwey:
When REBAR_DEPS_PREFER_LIBS is set, dependencies with defined sources
are allowed to be searched for in system lib directory.

Under specific circumstances (i.e. in build environments without networking)
it is impossible to fetch deps locally. So, user needs a way to ask rebar to
search in system lib directory as well.

Signed-off-by: Matwey V. Kornilov <matwey.kornilov@gmail.com>


lrascao:
+1, could you please append your name to the `THANKS` file, document the variable in the wiki (`Environment variables` section) and write a small test in `inttest/` to prevent future regressions? A test in the same style as the `tdeps/` would be sufficient

matwey:
REBAR_DEPS_FORCE_LIB
--------------------------------------

```REBAR_DEPS_FORCE_LIB``` is boolean environment variable. When it is set, rebar looks for all dependencies in system-wide lib directory.
In some restricted build environments all dependencies are usually already pre-installed into specific system-wide place, whereas internet access is prohibited which makes ```git clone``` useless.

tuncer:
We should explain REBAR_DEPS_FORCE_LIB in the rebar_deps help string (rebar_deps:info).

tuncer:
A short explanation can be something like this: "If REBAR_DEPS_FORCE_LIB is set, consider only deps found in the code path and skip fetching."

tuncer:
Thinking about it, would it be clearer to rename REBAR_DEPS_FORCE_LIB to REBAR_DEPS_FORCE_SYS, REBAR_DEPS_IGNORE_LOCAL or something similar. It's unclear to me what FORCE_LIB means.

matwey:
I think ```IGNORE_LOCAL``` is not exact because local is not ignored completely, it is just scanned after system.  ```REBAR_DEPS_FORCE_LIB``` is accordant with ```ERL_LIBS```
However, I am fine with any name.

tuncer:
How about PREFER_LIB or PREFER_SYS?

matwey:
Maybe ``` REBAR_DEPS_PREFER_LIBS``` in plural?

tuncer:
Works for me. @lrascao?

lrascao:
Sure, looks good.
@matwey could you please rebase onto master? we've merged some stuff in the meantime

matwey:
Any news? I don't remember whether you have email notifications when I rebase the request.

lrascao:
Sorry, got busy with stuff, will merge today

Issue #348, opened 2014-08-20
port_compiler: does the current is_expandable logic really make sense?


cfra:
The current `is_expandable` logic in `port_compiler` judges any variable that has a default definition containing other variables as expandable. At least to me, that behavior is a bit confusing. Take for example the following default definition for `ERL_LDFLAGS`:

```
ERL_LDFLAGS = -L$ERL_EI_LIBDIR -lerl_interface -lei
```

No  matter what I export for `ERL_LDFLAGS` when calling rebar, it will always be overridden by the default definition because `ERL_LDFLAGS` is deemed expandable. Is it really intentional that the user is unable to override the default definition for `ERL_LDFLAGS`?

My suggestion would be to only treat those variables as expandable which contain themselves in their default definition. An example for this would be `CFLAGS`:

```
CFLAGS = -m64 $CFLAGS
```

Here, imho, it makes much more sense to apply the default expansion to an existing export of `CFLAGS` than for the case of `ERL_LDFLAGS`.

bboozzoo:
@tuncer this will certainly fix cross-compilation, right now I cannot override `ERL_LDFLAGS` what results in native library paths being picked up instead of the target ones. As a workaround I'm setting ERL_EI_LIBDIR as it's used in the following context:

    ERL_LDFLAGS=-L$ERL_EI_LIBDIR -lerl_interface -lei'

and is itself non-expandable. Fortunately I can override `ERL_CFLAGS`, so in the end I don't have to ship a patched version of rebar.

lrascao:
@bboozzoo could you take a look at #590 and see if it fixes your use case? I would also appreciate feedback on some other edge case that we're not covering in the added test

lrascao:
c43f2e55 reversed this proposed fix due to a regression detected for already existing projects

lrascao:
Hi, this issue was closed in an attempt to do quick basic filtering. All tickets prior to March 2017 were closed and will be reopened on a per-request basis, please give rebar3 a try and feel free to reopen if it's a pressing issue that is still unresolved.

cfra:
What is the status then? For future reference it would be really useful if the way this was addressed was documented here in the issue. If it was not addressed, leaving the issue open would make more sense, in my humble opinion.

lrascao:
[...]

PR #590, opened 2016-02-19
Treat port env vars as expandable only if they self reference

lrascao:
Addresses #348 
Provide additional port test case.
Also, update newest OS X versions build flags.

### DIFF
--- a/inttest/port2/c_src/test1.c
+++ b/inttest/port2/c_src/test1.c
@@ -1,0 +1,13 @@
+#include "test1.h"
+
+#ifndef TEST1
+#error TEST1 is not defined!
+#endif
+
+#ifndef TEST2
+#error TEST2 is not defined!
+#endif
+
+#ifndef TEST3
+#error TEST3 is not defined!
+#endif
--- a/inttest/port2/port2_rt.erl
+++ b/inttest/port2/port2_rt.erl
@@ -1,0 +1,75 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil -*-
+%% ex: ts=4 sw=4 et
+%% -------------------------------------------------------------------
+%%
+%% rebar: Erlang Build Tools
+%%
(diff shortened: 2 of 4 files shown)


tuncer:
+1. This also resolves #182.

lrascao:
thanks, i'll look into it

lrascao:
@jflatow do you have a Makefile example with the issue you described? i'm having a bit of trouble replicating it

jflatow:
Sure, for example create a simple app called bar, with rebar.config:

```
{erl_opts, [debug_info, warnings_as_errors, warn_export_all]}.
{deps,
 [{'discodb', ".*",
   {git, "https://github.com/jflatow/discodb.git", "master"}, [raw]}]}.
{pre_hooks,
 [{compile, "make -C $REBAR_DEPS_DIR/discodb erlang"}]}.
```

And for completeness, src/bar.app.src:
```
{application, bar,
 [{vsn, []},
  {mod, {agency, []}},
  {applications, [kernel, stdlib]}]}.
```

lrascao:
I tried out your example but still can't reproduce it, can you paste the error you're getting and the output of `rebar --version` you're using? thanks!

jflatow:
On anything from b816c69e343c8fd757c59fe8703eeda597f4da0a...9281b848f563e4928b565fab57b610c7368c0be6 I get an error:

```
==> erlang (compile)
Compiling c_src/ddb.c
c_src/ddb.c:9:10: fatal error: 'discodb.h' file not found
#include <discodb.h>
         ^
1 error generated.
```

On 022afd4b08fe024520849e50fd2bd03294fe210a or earlier, it works.

lrascao:
got it, i see it now, i was using the 2.6.1 rebar that's in the path, that one doesn't have the regression

lrascao:
c43f2e55 revert fixes the regression

PR #592, opened 2016-02-25
Explicitly setting the git remote's URL when updating git dependencies

wk8:
Otherwise, switching remotes for a dependency (as often happens e.g. when forking
a github repo, then switching back again when the pull request gets merged by the
upstream repo's maintainers) would typically break builds, with errors like:

```
sh(git checkout -q <git_reference>)
failed with return code 128 and the following output:
fatal: reference is not a tree: <git_reference>
```


tuncer:
@wk8, I suppose you don't use rebar2 anymore, or else you would have pinged about the inactivity here. The diff looks reasonable to me and, if we are certain it won't negatively affect any scenario in the wild, I'd suggest a review+merge.

wk8:
@tuncer yup we did switch to rebar3 a few months ago, but used that version of rebar2 till then, up to you if you want to review/merge, happy to do changes if the review warrants them :)

tuncer:
In that case, do you mind extending the test suite? Given the potential to break scenarios in the wild, having tests for variations of git remote changes in deps is important. To that end, clone one of `inttest/tdep*` and execute tests by first fetching `retest` via `make deps` and then each running `deps/retest/retest -v inttest`. In the past, it was possible to selectively run a single inttest, but this has been broken since the introduction of a shared library helper module in `inttest`.

wk8:
@tuncer okay, will do soon, thanks

tuncer:
Thanks, maybe wait until #592 is merged or make sure you don't copy the `export_all` directive.

tuncer:
Thanks. Unless you want to wait for #641 to be merged first, make sure you don't copy the `export_all` directive when cloning tests.

tuncer:
Thinking about it, I haven't seen this reported anywhere else and, since you moved to rebar3, we could skip the patch and avoid regressions.

PR #593, opened 2016-03-02
No such function erlang:timestamp/0

lemenkov:
Hello All!
This looks like a typo for me - I can't find erlang:timestamp/0 anywhere, but I clearly see os:timestamp/0 available. So let's fix this.


tuncer:
```
1> erlang:timestamp().
{1456,922041,397711}
```

lemenkov:
@tuncer oh sorry - I didn't try 18.x.y.

PR #594, opened 2016-03-02
Introduce REBAR_VSN_CACHE_FILE env variable to load/save vsn cache

matwey:
When REBAR_VSN_CACHE_FILE is set, then vsn cache is loaded from
this file on rebar start and updated when new data.

Under specific circumstances (i.e. in build environments),
full git tree may not be available, but only its snapshot.
We need a way to use preheat vsn cache instead of invocing
git command.

Signed-off-by: Matwey V. Kornilov <matwey.kornilov@gmail.com>

### DIFF
--- a/inttest/vsn_cache/main.erl
+++ b/inttest/vsn_cache/main.erl
@@ -1,0 +1,13 @@
+-module(main).
+-behaviour(application).
+     
+-export([start/0,start/1,start/2,stop/1]).
+
+start() ->
+	start(permanent).
+start(_Restart) ->
+	ok.
+start(_Type,_Args) ->
+	ok.
+stop(_State) ->
+	ok.
--- a/inttest/vsn_cache/vsn_cache_rt.erl
+++ b/inttest/vsn_cache/vsn_cache_rt.erl
@@ -1,0 +1,90 @@
+%% -*- erlang-indent-level: 4;indent-tabs-mode: nil -*-
+%% ex: ts=4 sw=4 et
+-module(vsn_cache_rt).
+
+-compile(export_all).
+
+setup([Target]) ->
+    retest_utils:load_module(filename:join(Target, "inttest_utils.erl")),
+    ok.
+
+files() ->
+    [
+    %% Cache save check
+    {create, "save/src/save.app.src", app(save, [main])},
+    {create, "save/vsn_cache_file", ""},
+    {copy, "main.erl", "save/src/main.erl"},
+
+    %% Cache load check
+    {create, "load/src/load.app.src", app(load, [main])},
+    {copy, "main.erl", "load/src/main.erl"}
+    ] ++ inttest_utils:rebar_setup().
+
+apply_cmds([], _Params) ->
+    ok;
+apply_cmds([Cmd | Rest], Params) ->
+    io:format("Running: ~s (~p)\n", [Cmd, Params]),
+    {ok, _} = retest_sh:run(Cmd, Params),
+    apply_cmds(Rest, Params).
+
+run_save(Dir) ->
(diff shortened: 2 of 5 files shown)


lrascao:
+1, could you please document the variable in the wiki (`Environment variables` section) and write a small test in `inttest/` to prevent future regressions?

matwey:
@lrascao should I squash tests with this commit?

matwey:
I am afraid I don't have access to rebar/rebar wiki to edit ```Environment variables``` section

lrascao:
Please do

On Saturday, 12 March 2016, Matwey V. Kornilov <notifications@github.com>
wrote:

> @lrascao <https://github.com/lrascao> should I squash tests with this
> commit?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rebar/rebar/pull/594#issuecomment-195746823>.
>


-- 
PGP fingerprint: F708 E141 AE8D 2D38 E1BC  DF3D 1719 3EA0 647D 7260

lrascao:
Ok, then please just post the text here and i'll put it up on the wiki

On Saturday, 12 March 2016, Matwey V. Kornilov <notifications@github.com>
wrote:

> I am afraid I don't have access to rebar/rebar wiki to edit Environment
> variables section
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rebar/rebar/pull/594#issuecomment-195749723>.
>


-- 
PGP fingerprint: F708 E141 AE8D 2D38 E1BC  DF3D 1719 3EA0 647D 7260

matwey:
REBAR_VSN_CACHE_FILE
-----------------------------------

When ```REBAR_VSN_CACHE_FILE``` is set, it should point to existing, readable and writable file where vsn cache is stored. vsn cache is a key-value storage where results of ```git describe --always --tags``` are stored. Rebar always uses in-memory vsn cache internally. ```REBAR_VSN_CACHE_FILE``` provides a way to make the cache persist between different ```rebar``` invocations which may reduce number of ```git``` invocations.
Moreover, ```REBAR_VSN_CACHE_FILE``` can be used with externally prepared cache file when you have only a snapshot of git repository and ```.app.src``` file  has ```{vsn, git}``` term, in this case ```git``` invocation on non-git directory tree will be avoided.

tuncer:
How long do you expect the cache file to exist? Would we need to handle conflicts with multiple projects?

matwey:
Rebar uses global absolute paths in cache storage, so there are no conflicts between different instances of the same repo. The single way to make the cache file inconsistent is to do something like ```git checkout```, ```git commit```, ```git pull```, etc. If the work flow includes such operations, then cache file should not be used. The cache file is enabled explicitly via environment variable., so there is no chance to break something accidentally.
I expect that the cache file would be per-project or per-build and used in scripted non-interactive environments. So, I imagine generalized build scenario (like CI script, or RPM spec-file as in my case). Something like the following (pseudo-bash):
```
#!/bin/sh

# Set cache
export REBAR_VSN_CACHE_FILE=vsn_cache
# Fill cache
echo "{{git, \"/build/theproject\"},\"1.0.0-abcdef-build123\"}." > vsn_cache
## Or just make empty one
#touch vsn_cache
# Untar something
tar -zxf ...
# Build something
rebar compile
# Collect artifacts
tar -zcf ...
# Deploy artifacts
scp ...
# Cleanup
rm -rf /build/theproject
rm vsn_cache
```

tuncer:
Okay. Let's mention the env var together with a sample entry in the documentation, and it should also be included in the common help string (rebar:help/0).

matwey:
I've rebased this on master.

lrascao:
thanks!

PR #595, opened 2016-03-06
Fix bug when running gcc in cross_sizeof

fhunleth:
The test program used to determine the word size of a crosscompiler is
crafted to return an error, so this changes the logic to expect an
error exit. If the crosscompiler actually compiles the test program
successfully, that would be remarkable and worthy of investigation.

To test, download the i586 crosscompiler from:
(If on OSX) https://github.com/nerves-project/nerves-toolchain/releases/download/v0.6.0/nerves-i586-unknown-linux-gnu-darwin-x86_64-v0.6.0.tar.xz
(If on Linux) https://github.com/nerves-project/nerves-toolchain/releases/download/v0.6.0/nerves-i586-unknown-linux-gnu-linux-x86_64-v0.6.0.tar.xz

Untar the toolchain and put the `bin` directory in your `PATH`.

`export REBAR_TARGET_ARCH=i586-unknown-linux-gnu`

Run `rebar compile` on any project.

### DIFF
--- a/src/rebar_utils.erl
+++ b/src/rebar_utils.erl
@@ -760,7 +760,7 @@ cross_sizeof(Arch, Type) ->
                          >>),
     Cmd = Compiler ++ " -DTYPE=\""++Type++"\" " ++ TempFile,
     ShOpts = [{use_stdout, false}, return_on_error],
-    {ok, Res} = sh(Cmd, ShOpts),
+    {error, {_,Res}} = sh(Cmd, ShOpts),
     ok = file:delete(TempFile),
     case string:tokens(Res, ":") of
         [_, Ln | _] ->


tuncer:
+1. Can you add tests (in `inttest/`) and also append your name in THANKS
(as part of the original commit)?

tuncer:
+1. Can you add tests (in `inttest/`) and also append your name in THANKS (as part of the original commit)?

fhunleth:
Hmm. I'm not sure how to test `cross_sizeof` as an `inttest`. Sorry - I'm not a `rebar` expert by any stretch.

lrascao:
@fhunleth do you think there might be a way of devising a test that doesn't involve download an external resource (like the nerves compiler you mentioned)?

fhunleth:
The code triggers on the name of the compiler, so if you symlinked `coolprocessor-unknown-linux-gnu-gcc` to the system `gcc`, exported `REBAR_TARGET_ARCH=coolprocessor-unknown-linux-gnu` and made sure that the symlink was in the PATH, I'm pretty sure that you could see the problem. I haven't verified that it works, but I really think it should.

lrascao:
@fhunleth I've submitted #599 which includes your fix and a simple regression test, it's capturing the issue and ceases to fail with your fix applied, if it's fine by you i'll close this issue in favor of #599

fhunleth:
@lrascao Thanks for writing the regression test! Yes, please close this issue in favor of #599.

PR #598, opened 2016-03-12
Fix Eunit tests with recent Erlang 18.x.y

lemenkov:
In a test-suite contains more than one test, then Eunit prints "All \<n\>
tests passed". The only exception is the case then a test-suite contains
only two tests. In this case Eunit outputs sligntly different message -
"2 tests passed".

The original Eunit behaviour was changed in this commit -
erlang/otp@63d9357482d977f110c1b1804d827bb81d513cfa.

Signed-off-by: Peter Lemenkov <lemenkov@gmail.com>


lemenkov:
Sorry for the noise - this was fixed in master branch in ea84fdaf31fddab55ba3d9630710b21e5db209bc already.

PR #602, opened 2016-03-19
Support full-source rebar3-style deps (deps without Version Regexes)

choptastic:
In order to support applications with rebar3, full-path dependencies like `{git, {appname, "git://something/something", {branch, master}}}` (That is, the source without the Version Regex).

As an example that this patch allows to compile with rebar2 is erlpass https://github.com/ferd/erlpass/blob/master/rebar.config#L2

(at least, this is something to try to smooth things over while projects get around to upgrading to rebar3)

Thoughts?


tuncer:
Can you add a test case in `inttest/`? Other than that, this seems reasonable to me. @lrascao?

lrascao:
+1, also please add your name to 'THANKS', if not already there

On Sunday, 20 March 2016, Tuncer Ayaz <notifications@github.com> wrote:

> Can you add a test case in inttest/? Other than that, this seems
> reasonable to me. @lrascao <https://github.com/lrascao>?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/rebar/rebar/pull/602#issuecomment-198931888>
>


-- 
PGP fingerprint: F708 E141 AE8D 2D38 E1BC  DF3D 1719 3EA0 647D 7260

choptastic:
I'll get a test case in there today. Thanks!

choptastic:
Test case added (called `rebar3_deps1`) and commits squashed.

tuncer:
Shouldn't we also extend existing docs?

choptastic:
Definitely. Once merged, I can make the appropriate changes in https://github.com/rebar/rebar/wiki/Dependency-management

tuncer:
You also have to extend `rebar_deps:info_help/1`.

choptastic:
Ah, right. Forgot about the built-in help.

choptastic:
I realized my fix leaves off support for VsnRegex-Free raw dependencies. In the name of consistency, I'm thinking it would make sense to add support there as well, basically making VsnRegex optional across the board.

Any objections?

choptastic:
Updated, though I kept the raw support as a separate commit.  Docs updated in each and a test case added for raw.

choptastic:
Do my documentation changes (and the additional raw support) look acceptable?  For the docs changes, you can see I merely offered all the current rules except without the VsnRegex - not sure if that's a little *too much* by way of verbosity, but it does certainly drive the point home.

choptastic:
Thoughts, comments, objections?

lrascao:
Looks good to me, @tuncer?

tuncer:
On 9 April 2016 at 16:08, Luis Rascão <notifications@github.com> wrote:
>
> Looks good to me, @tuncer?

+1

choptastic:
Thanks fellas!

--
Jesse Gumm
Owner, Sigma Star Systems
414.940.4866 || sigma-star.com || @jessegumm

PR #603, opened 2016-04-13
add deprecated message

tsloughter:


ferd:
I'm okay with this.

tsloughter:
A message should be added to the github description as well I think.

PR #611, opened 2016-07-17
port_compiler: consistently format default_env/0

tuncer:


tuncer:
Would be great if we can merge this trivial fix before the attempt to sync port_compiler with the rebar3 plugin.

PR #612, opened 2016-07-25
Introduce and use compat random number module

tuncer:
In order for rebar to work with previous and current OTP releases, we
generate the rebar_rnd module on startup. rebar_rnd is generated
dynamically to make sure that we use the right random module (either
rand if available or else random). It only wraps the common subset of
the API, but that's sufficient for rebar's use.

### DIFF
--- a/ebin/rebar.app
+++ b/ebin/rebar.app
@@ -46,7 +46,8 @@
               rebar_metacmds,
               rebar_getopt,
               rebar_mustache,
-              rmemo ]},
+              rmemo,
+              rebar_rand_compat ]},
   {registered, []},
   {applications,
    [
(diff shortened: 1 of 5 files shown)


tuncer:
Tested with 18.3.3 and 19.0.2.

jj1bdx:
@tuncer on `rand_compat` module: it looks OK. One suggestion: if you do not need to repeat the same seeding every time, `rand:seed(exsplus)` is enough and actually does *better* than `rand:seed(exsplus, {42, 43, 44})`, because as default rand:seed/1 will pick up the primary seeding data from internal time functions (and will not repeat the same default seeding value, which no longer exists in the rand module). See https://github.com/erlang/otp/blob/OTP-19.0.2/lib/stdlib/src/rand.erl#L86L93 for the details.

tuncer:
@jj1bdx, yeah, that could make sense, but let's get the compat fix in and tackle that later.

tuncer:
Dropped 19.0 enablement patch until OTP regression is fixed.

PR #615, opened 2016-08-18
port_compiler: generate clang compilation db

tuncer:
In order for newer clang tools to work, they require the presence of a
compilation database in the form of compile_commands.json.  Therefore,
we adapt port_compiler to write such a file.


tuncer:
More info: http://eli.thegreenplace.net/2014/05/21/compilation-databases-for-clang-based-tools

PR #617, opened 2016-08-26
port_compiler: auto-select C++ specific link template

tuncer:
Even though "$CC foo.o -lstdc++ -o foo.so" works if foo.o was built with
$CXX, depending on the platform and toolchain and C++ standard used, it
may not. In order to ensure this works reliably, we introduce
EXE_LINK_CXX_TEMPLATE and DRV_LINK_CXX_TEMPLATE and auto-select if the
port spec being built consists of at least one C++ source.

### DIFF
--- a/src/rebar_port_compiler.erl
+++ b/src/rebar_port_compiler.erl
@@ -36,10 +36,11 @@
 -include("rebar.hrl").
 
 -record(spec, {type::'drv' | 'exe',
+               link_lang::'cc' | 'cxx',
                target::file:filename(),
                sources = [] :: [file:filename(), ...],
                objects = [] :: [file:filename(), ...],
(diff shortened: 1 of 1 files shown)


tuncer:
@laverite, can you verify this fixes https://github.com/tuncer/re2/issues/15?

tuncer:
Alternatively, we could introduce `$LINKER` and used that instead of `$CC` in `$EXE_LINK_TEMPLATE` and `$DRV_LINK_TEMPLATE`.

laverite:
@tuncer thank you, yeah that works for me when I use rebar.  but i'm using rebar3 :D.

tuncer:
@lrascao, can you please review?

lrascao:
@tuncer looks good, merging

PR #619, opened 2016-08-28
Fix cdb processing when a file is skipped

tuncer:
This fixes a bug introduced in the new code in 2.6.3. We should release 2.6.4.


lrascao:
Are there no regression tests for this?

tuncer:
None that would have caught this. I've noticed this during the backport of changes to blt/port_compiler.

tuncer:
I'll extract that into a variable for good measure.

tuncer:
Investigating a build failure in https://github.com/travelping/gen_socket.

tuncer:
Fixed handling of multiple specs.

PR #620, opened 2016-08-30
Travis dialyze

ghost:
There are new Dialyzer errors. To not introduce any new warnings, try to enable Dialyzer on Travis-CI.


tuncer:
Added GitLab CI config file. Here's the GitLab mirror CI: https://gitlab.com/rebar/rebar-tuncer/pipelines

tuncer:
Preliminary results: GitLab CI doesn't timeout and manages to create the PLT. Switched default otp to 18 so that Dialyzer can actually succeed.

tuncer:
bump?

tuncer:
Some workers timeout, others succeed and report Dialyzer warnings. When this is merged, I'd suggest to manually clear the Travis-CI caches once (as described under `Clearing Caches` in https://docs.travis-ci.com/user/caching/).

tuncer:
Given the timesharing constraints of Travis-CI, I should probably rework this to have one job that only builds and dialyzes rebar, saving the time spent in eunit and retest for Dialyzer. Need to check .travis-ci.yml docs to see how this can be achieved.

tuncer:
Until all warnings (#621) have been fixed, I will probably have to turn it into a report-only CI job, not affecting the success code.

tuncer:
Reading https://docs.travis-ci.com/user/customizing-the-build, but still not sure how to define a matrix that runs `make travis-dialyze` separately and gives Dialyzer more cpu time.

tuncer:
Finally figured out how to structure travis config and the results look good. One inttest CI timeout on 18 and as expected errors in the separate Dialyzer job. If we don't merge this, we will only accumulate more Dialyzer errors.

tuncer:
Waiting for #643 to be merged.

tuncer:
With the two silenced opaque match warnings, we're down to this:

```
src/rebar_config.erl:103: Record construction
#config{dir::'undefined',opts::[any()],globals::dict:dict(_,_),envs::dict:dict(_,_),skip_dirs::dict:dict(_,_),xconf::dict:dict(_,_)}
violates the declared type of field dir::string()
src/rebar_core.erl:334: The call
rebar_config:new(ParentConfig::rebar_config:config()) contains an
opaque term as 1st argument when a structured term of type string() is
expected
```
I want to say when calling rebar_config:new, it should also set dir to CWD, but not sure.

---

src/rebar_erlc_compiler.erl:543: Function compile_mib/3 has no local return
src/rebar_erlc_compiler.erl:554: Record construction
#options{includes::[],outdir::[46,...],defines::[],warning::1,verbose::'false',optimize::999,specific::[],outfile::[],cwd::'undefined'}
violates the declared type of field cwd::string()
src/rebar_erlc_compiler.erl:556: Record construction
#options{includes::[],outdir::[46,...],defines::[],warning::1,verbose::'false',optimize::999,specific::[{'verbosity',_},...],outfile::[],cwd::'undefined'}
violates the declared type of field cwd::string()

Not sure. Maybe related to the rebar_config warning above?

---

src/rebar_xref.erl:261: The pattern {Source, 'function_not_found'} can
never match since previous clauses completely covered the type
{'module_not_found','function_not_found'}
src/rebar_xref.erl:262: The pattern {Source, Line} can never match
since previous clauses completely covered the type
{'module_not_found','function_not_found'}
src/rebar_xref.erl:268: Function safe_element/2 will never be called
src/rebar_xref.erl:287: Function find_function_source/4 has no local return

Tried to fix this, but couldn't figure it out.

tuncer:
@lrascao, maybe Dialyzer is more capable in 20.0 and has revealed issues that are quite old. I mean, that discrepancy in `rebar_config` looks risky to fix by overriding/initializing the dir with CWD. Any thoughts on whether we should instead try to silence the warning, or do you have an idea how to resolve it?

lrascao:
The only idea i have is to use cwd, but as you say, it's probably not worth the risk, if there's a way to instead silence the warning that would be preferable

tuncer:
Yeah, it might be a genuine bug, but explicitly setting dir to cwd in that function, which hasn't been touched in ages, is very risky indeed. The field in the record hasn't been touched since 2011 and the function (Dialyzer reports) was last modified 2012. Let's try to silence it.

tuncer:
In order to get this merged and not get lost, we now allow the Dialyzer job to fail.

PR #632, opened 2017-02-21
Fix typo in eunit help.

phiggins:
I'm an erlang newbie and was briefly confused by a small typo in the help documentation for `./rebar help eunit`.


lrascao:
Thanks!

PR #637, opened 2017-08-18
Drop crypto app reliance

tuncer:
By doing this, we fix two issues at once. First, there's no need to have
crypto available anymore. While not having crypto in your Erlang
installation is a questionable packaging decision, it tends to happen in
the wild with users installing Erlang and missing crypto. Sometimes this
is not due to a distro's package but users building Erlang without the
needed OpenSSL dependencies. Second, this resolves the Erlang 20 rng
deprecation warnings.

### DIFF
--- a/ebin/rebar.app
+++ b/ebin/rebar.app
@@ -55,7 +55,6 @@
     stdlib,
     sasl,
     compiler,
-    crypto,
     syntax_tools,
     tools,
     eunit,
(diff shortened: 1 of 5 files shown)


tuncer:
Replacing `crypto:rand_uniform(0, 10000)` with `rebar_rnd:uniform(10000)` (which is either `rand:uniform` or `random:uniform` dependent on what otp version it runs on) fails. I suspect it's because the old call requests a number `>= 0` while the new call has no way to ask for anything but `>= 1`.

tuncer:
Mirroring what `rebar_eunit` does and using a bigger rng range seems to be enough to avoid node name clashes on Travis-CI.

jj1bdx:
I have no objection on using rebar_rnd:uniform/1 instead of crypto:rand_uniform/2.

PR #638, opened 2017-08-18
Fix Erlang 20 export_all warnings

tuncer:
Fix other warnings and whitespace errors as well.


tuncer:
2 of the 10 CI jobs have timed out, but otherwise succeeded.

tuncer:
There are also `export_all` occurrences in `inttest/` and in the `priv/templates/ctsuite.erl` template.

PR #639, opened 2017-08-18
rebar.config: access retest repo via https

tuncer:


tuncer:
One CI job failed with a timeout.

PR #640, opened 2017-08-19
Fix ctsuite template

tuncer:
- Fix and test the Erlang code.
- Do not export_all.


tuncer:
One CI job timed out.

Issue #608, opened 2016-06-28
-p profiler=PROFILER TODO (dependent on new getopt feature)


tuncer:
`rebar -p profiler=fprof/eflame` shall be replaced with `rebar -p fprof`, while still allowing `rebar -p` to mean "profile with default profiler". For this to work, we must first implement https://github.com/jcomellas/getopt/issues/34.

PR #642, opened 2017-08-19
Remove profiler note

tuncer:
If and when getopt gains the required feature, this can be revisited,
but let's remove the message, since it's been a long time.


tuncer:
Related: #608.

PR #643, opened 2017-08-20
Revert "Drop crypto app reliance"

tuncer:
Reverts rebar/rebar#637

### DIFF
--- a/ebin/rebar.app
+++ b/ebin/rebar.app
@@ -55,6 +55,7 @@
     stdlib,
     sasl,
     compiler,
+    crypto,
     syntax_tools,
     tools,
     eunit,
--- a/rebar.config
+++ b/rebar.config
@@ -32,8 +32,7 @@
       - (\"diameter_dict_util\":\"parse\"/\"2\")
       - (\"erlang\":\"timestamp\"/\"0\")
       - (\"rebar_rnd\":\"seed\"/\"1\")
-      - (\"rebar_rnd\":\"uniform\"/\"0\")
-      - (\"rebar_rnd\":\"uniform\"/\"1\"))",
+      - (\"rebar_rnd\":\"uniform\"/\"0\"))",
          []}]}.
 
 {dialyzer,
--- a/src/rebar.erl
+++ b/src/rebar.erl
@@ -205,6 +205,12 @@ profile(_Config, _Commands, Profiler) ->
     ?ABORT("Unsupported profiler: ~s~n", [Profiler]).
 
 run_aux(BaseConfig, Commands) ->
+    %% Make sure crypto is running
+    case crypto:start() of
+        ok -> ok;
+        {error,{already_started,crypto}} -> ok
+    end,
+
     %% Make sure memoization server is running
     case rmemo:start() of
         {ok, _} -> ok;
--- a/src/rebar_ct.erl
+++ b/src/rebar_ct.erl
@@ -237,7 +237,6 @@ make_cmd(TestDir, RawLogDir, Config) ->
     CodeDirs = [io_lib:format("\"~s\"", [Dir]) ||
                    Dir <- [EbinDir|NonLibCodeDirs]],
     CodePathString = string:join(CodeDirs, " "),
-    _ = rebar_rnd:seed({55, seconds(), 7331}),
     Cmd = case get_ct_specs(Config, search_ct_specs_from(Cwd, TestDir, Config)) of
               undefined ->
                   ?FMT("~s"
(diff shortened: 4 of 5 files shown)


tuncer:
Revert, since Travis still can fail (https://travis-ci.org/rebar/rebar/jobs/266578654 in https://travis-ci.org/rebar/rebar/builds/266578650).

jj1bdx:
rand module since OTP 20 depends on the crypto module, so I think rebar.app should include crypto.

tuncer:
So the dynamic compat module is fine, but we should retain the `crypto:start()` call. Or does `rand` start crypto on demand?

jj1bdx:
I might be wrong. It's vice versa: crypto now depends on rand. Check the manual here
http://erlang.org/doc/man/crypto.html#rand_seed-0
(I'm not really familiar with the dependency between crypto and rand since OTP 20 yet)

rand does not start crypto at all. See
https://github.com/erlang/otp/blob/master/lib/stdlib/src/rand.erl

tuncer:
I missed the deprecation change in 20.0 and must say it's interesting that crypto is needed for rand:uniform but crypto:rand_uniform will be removed in the future.

tuncer:
I see, so while rand:uniform does not call into crypto, we are expected to use the seeder from crypto. But then I'm confused why we have rand:seed. I mean, when I use random:uniform or rand:uniform, the docs don't indicate that I should not use rand:seed/random:seed. Also, the docs for crypto:rand_seed mention it being for rand:uniform use.

jj1bdx:
rand:seed is *implicitly* executed at the first invocation of rand:uniform. rand:seed's result will not be fixed at all (depending on all the timers and process ids are other variable factors in the BEAM) so this will be ok.

OTOH random:uniform without rand:seed results in a *fixed* result of seeding (by the *implicit* execution), so random:uniform *must* be explicitly seeded with a variable factor to obtain different results.

I believe some people developing crypto module thought the current plugin interface of rand module was suitable to encapsulate cryptographically-strong random number generation functions for OTP 20. I haven't really checked out the crypto module yet, but I guess some functions in crypto module are actually forming plugins of rand module.

tuncer:
@lrascao, let's merge this revert and re-introduce the 20.0.2 warnings. We can try to fix this again, probably differently, and a warning is better than spurious Travis-CI errors in the ct_run test suite.

tuncer:
On 8/21/17, Kenji Rikitake <notifications@github.com> wrote:
> @tuncer: did you get the deprecation warning for `crypto:rand_uniform/2`? I
> guess this code in `otp_internal.erl` generates it:
> https://github.com/erlang/otp/blob/OTP-20.0.2/lib/stdlib/src/otp_internal.erl#L107L111

Yes. I'll come up with a new patch to approach this differently in an
hour or so, and we can continue the discussion there.

tuncer:
@jj1bdx, yes that's the warning.

tuncer:
tl;dr: I don't know how to fix the warning, especially in a backwards compatible (OTP version and rebar features) way.

Long story: I was thinking we could seed in rebar:main and call rebar_rnd:uniform in rebar_ct and rebar_eunit, but it's a more complicated. The EUnit tests randomization/?shuffling? feature looks for input from the PRNG seed on the command line. I don't know how we would keep that and centralize PRNG initialization.

tuncer:
tl;dr: I don't know how to fix the warning, especially in a backwards compatible (OTP version and rebar features) way.

Long story: I was thinking we could seed in rebar:main and call rebar_rnd:uniform in rebar_ct and rebar_eunit, but it's more complicated. The EUnit tests randomization/?shuffling? feature looks for the PRNG seed on the command line. I don't know how we would keep that and centralize PRNG initialization.

Here's what I think the PRGN use was intended to achieve:

- rebar_ct: create a unique nodename for use as the node when spawning `ct_run`
 - Like I said before, a PRNG isn't really the right solution here anyway.
- rebar_eunit: shuffle tests, but I could be wrong.
 - A shuffling algo like Fisher-Yates appears more application here.

Both PRGN uses seem to not require PRNG (and crypto) at all, if solved in a fundamentally different and arguably more appropriate fashion.

tuncer:
tl;dr: I don't know how to fix the warning, especially in a backwards compatible (OTP version and rebar features) way.

Long story: I was thinking we could seed in rebar:main and call rebar_rnd:uniform in rebar_ct and rebar_eunit, but it's more complicated. The EUnit tests randomization/?shuffling? feature looks for the PRNG seed on the command line. I don't know how we would keep that and centralize PRNG initialization.

Here's what I think the PRNG use was intended to achieve:

- rebar_ct: create a unique nodename for use as the node when spawning `ct_run`
  - Like I said before, a PRNG isn't really the right solution here anyway.
- rebar_eunit: shuffle tests, but I could be wrong.
  - A shuffling algo like Fisher-Yates appears more applicable here.

Both PRGN uses seem to not require PRNG (and crypto) at all, if solved in a fundamentally different and arguably more appropriate fashion.

RaimoNiskanen:
@tuncer pulled me in here, so I am not sure I have caught up with everything..., but I will try to throw some random hints at this.

@jj1bdx  is right as far as I can see.  The `rand` module does not depend on `crypto`, but the `crypto application calls the `rand` module from `crypto:rand_seed/0` and `crypto:rand_seed_s/0`.  I guess you are not interested in cryptographically strong numbers here, so the `crypto` application should not be needed.

You can suppress warnings, for example by creating a wrapper module that calls all functions you do not want deprecation warnings about.  Then you turn off all deprecation warnings in that wrapper module.  Your wrapper module can use `try` - `catch` to find a useful implementation.

If seeding is a problem you can re-implement a seed function yourself for example by mimicking the code in the last clause of OTP-20.0 `rand:seed_s/1`.  You only need to produce a 3-tuple of non-negative integers with a quality of your own choice to seed `rand:seed/2`, `rand:seed_s/2` or `random:seed/1`.

PR #645, opened 2018-01-29
Log out success message with newlines

Juliusan:
The raw log file is split into lines to check for errors.
However the logged message should contain newlines for easier reading.


lrascao:
I'm ok with merging this, however there are no more releases coming out of this project as it's been deprecated and replaced with [rebar3](https://www.rebar3.org/)

Juliusan:
That's ok with me. At least this will be in the main project and not just in my computer.

